PR #590, opened 2013-06-28
[on hold] Let more appenders survive on Aria.verbose

piuccio:
When `Aria.verbose = true` we remove all appenders and add back the `DefaultAppender`.
With this fix we keep all appenders already there, this allows other appenders like `WindowAppender` to keep working while testing


piuccio:
I don't think this will ever be merged

PR #1049, opened 2014-03-24
[on hold] fix #1049 Improve JsonValidator errors logging

jakub-g:
JsonValidator raises error in a special manner, by throwing an instance of
`Error` with `errors` array property (because there can be multiple validation
failures per one run).

This exception was not taken into account, hence errors from JsonValidator
were logged with empty error message in the test runner.

---

To see the minimal diff, append `?w=1` to URL


divdavem:
I am closing this PR, as it is superseded by #1740

Issue #74, opened 2012-07-18
html:Template widget data model bug


flongo:
Data model not correctly set on sub-templates associated to a moduleCtrl when using a html:Template widget

PR #1370, opened 2014-12-05
fix "Assert #74 failed in aria.core.Timer" in test launcher

divdavem:


test/testLauncher.js
```diff
@@ -13,6 +13,7 @@ var endTest = function (evt) {
 
     // Show information about objects created/disposed
     aria.core.Timer.addCallback({
+        scope : Aria,
```
divdavem: In fact, when there is no scope and no arg (as in this case), we could use a simple call to `setTimeout`.

divdavem:
I have updated this PR to use `setTimeout` instead of `aria.core.Timer.addCallback`.
As the indentation changed, it is better to look at [this diff which ignores white space changes](https://github.com/ariatemplates/ariatemplates/pull/1370/files?w=0)

jakub-g:
26de55b8d25901ed3296ff80e2168d61f0835e20 LGTM

PR #1371, opened 2014-12-05
Fixing issue with recursive dependency on VisualFocus

divdavem:
This commit fixes the following bug: if `appOutlineStyle` is set to a non-null value in the `AppEnvironment` (after Aria Templates is loaded but before any template is loaded), and if the packaging of Aria Templates is such that `aria.utils.VisualFocus` is available synchronously when `aria.utils.environment.VisualFocus` is loaded, `aria.utils.VisualFocus` does not get a correct reference to `aria.utils.environment.VisualFocus` when it is loaded, leading to the `undefined is not a function` exception when executing `ariaUtilsEnvironmentVisualFocus.getAppOutlineStyle()` in https://github.com/ariatemplates/ariatemplates/blob/master/src/aria/utils/VisualFocus.js#L39

### DIFF
--- a/src/aria/utils/environment/VisualFocus.js
+++ b/src/aria/utils/environment/VisualFocus.js
@@ -13,9 +13,20 @@
  * limitations under the License.
  */
 var Aria = require("../../Aria");
+var Promise = require("noder-js/promise");
+var asyncRequire = require("noder-js/asyncRequire");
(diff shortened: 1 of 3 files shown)


simonarbuckle:
:white_check_mark: 22278a7 is fine for me.

PR #1377, opened 2014-12-16
fix #1377 loading overlay juddering on scroll in Safari Mobile

jakub-g:
This PR changes default overlay position from fixed to absolute to prevent an issue on Safari Mobile where fixed-positioned overlay is not repainted on screen during scrolling until the scrolling ends.

I've tested the fix on the requesting application and on the following sample, on ipad (Safari 8), galaxy tab with Chrome, and on desktop browsers to check for regressions (Firefox, Chrome, IE7-9)

* http://ariatemplates.com/samples/?path=samples.templates.domInteractions.processingIndicator.Sample

I also tested those samples in major desktop browsers (Fx, Ch, IE7-9) for non-regression:

* http://ariatemplates.com/samples/?path=samples.widgets.dialog.resizable.TemplateResizableDialog
* http://ariatemplates.com/samples/?path=samples.widgets.dialog.movable.TemplateMovableDialog

### DIFF
--- a/src/aria/utils/Dom.js
+++ b/src/aria/utils/Dom.js
@@ -604,7 +604,8 @@ module.exports = Aria.classDefinition({
         },
 
         /**
-         * Get the absolute position in a page of an element
+         * Get the position in a page of an element. It is relative to the current viewport (i.e. depends on scroll of
+         * the page)
          * @param {HTMLElement} element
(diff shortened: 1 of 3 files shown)


jakub-g:
This didn't land in 1.7.1 as it was found to cause regressions during release testing.

PR #1380, opened 2015-01-06
Escape html in @aria:Textarea widget

flongo:
If the data model contains, for example, "&pound" the field should display it as it is. Before this commit the content of the text area would be a pound sign.


divdavem:
:ok: df5265fa7a02684f3591faa566abec42e6d58939 seems fine for me

PR #1382, opened 2015-01-06
feat #1382 logWarn when adding callback with undefined scope

jakub-g:
We used to log a cryptic error via assert


jakub-g:
related to #1370

divdavem:
:ok: 4bb74b3dde4dc24b4a240c94ad4b47669de69866 seems fine for me.

PR #1384, opened 2015-01-07
Improving the stability of the test suite

divdavem:
Some tests are randomly failing when run on some browsers. This commit partly fixes those kinds of failures.

### DIFF
--- a/test/aria/widgets/form/multiautocomplete/BaseMultiAutoCompleteTestCase.js
+++ b/test/aria/widgets/form/multiautocomplete/BaseMultiAutoCompleteTestCase.js
@@ -47,7 +47,11 @@ Aria.classDefinition({
                 text = [text];
             }
             this.synEvent.click(this._getField(), {
-                fn : this.type,
+                fn : function (evt, args) {
+                    this.waitForWidgetFocus("MultiAutoId", function () {
+                        this.type(args);
+                    });
+                },
                 scope : this,
                 args : {
                     text : text,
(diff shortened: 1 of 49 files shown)


test/aria/widgets/form/multiautocomplete/BaseMultiAutoCompleteTestCase.js
```diff
@@ -59,11 +63,24 @@ Aria.classDefinition({
...
+
+            if (typeof(nextText) == "function") {
+                this.waitFor({
+                    condition : nextText,
+                    callback : cb
+                });
+            } else {
+                // nextText is a string
+                this.synEvent.type(this.getFocusedElement(), nextText, cb);
+            }
```
jakub-g: I've realized there might be some kind of confusion happening here:

The `clickAndType` method here overrides `clickAndType` from TemplateTestCase [2] (since this class extends RobotTestCase which extends TemplateTestCase)

Hence adding the feature of accepting callback parameters will only work for MAC tests, not all Template tests. It would probably be interesting to move this feature to the TTC class and refactor the code of  `clickAndType` here and in the parent class accordingly.

[1] https://github.com/divdavem/ariatemplates/blob/randomlyFailingTests/test/aria/widgets/form/multiautocomplete/BaseMultiAutoCompleteTestCase.js#L45-84
[2] https://github.com/divdavem/ariatemplates/blob/randomlyFailingTests/src/aria/jsunit/TemplateTestCase.js#L702-725

jakub-g:
Apart from the earlier comment on `clickAndType` this LGTM

PR #1385, opened 2015-01-08
Blur after destroying a TextInput

flongo:
In IE, after destroying a TextInput having focus, the focus was not correctly released.
This had an impact on the behaviour of the application.

For example, when clicking on a button after the refresh operation, the click event was not being raised.


divdavem:
:ok: 9551dfeecf1c6f210a0e12be21c5240877958b58 seems fine for me.

Issue #1299, opened 2014-09-17
`aria.core.Browser` and `aria.utils.Device` properties are not correct


lsimone:
As shown in the image, many of the `aria.utils.Device` properties are not correct on Nexus 10 (Chrome browser)

![20140917_161521](https://cloud.githubusercontent.com/assets/3266650/4307198/65ac1276-3e87-11e4-9c40-a7522a6d0d3b.jpg)

PR #1386, opened 2015-01-09
Loading Overlay for Safari once again (obsoletes #1377)

jakub-g:
Fixed a test failure introduced in the commit 2aa83bb279c1f10d11fca7ee46e52869b3c4ee67 (PR #1377) causing the loading overlay not to be displayed when page is scrolled too far and some other related issues.

### DIFF
--- a/src/aria/utils/overlay/Overlay.js
+++ b/src/aria/utils/overlay/Overlay.js
@@ -70,7 +70,16 @@ module.exports = Aria.classDefinition({
                 overlay.id = "xOverlay" + params.id;
             }
             overlay.className = params.className || "xOverlay";
-            overlay.style.position = params.position || "fixed";
+
+            // absolute positioning is preferred over fixed since Safari Mobile doesn't repaint fixed elements on scroll
+            var position = params.position || "absolute";
+
+            // Place it in the top-left corner of the page, otherwise if it happens to be inserted far in the DOM, it
+            // may temporarily create unnecessary scrollbars on the page - this may affect proper positioning of the
+            // overlay. We'll set left/top later.
+            overlay.style.left = 0;
+            overlay.style.top = 0;
+            overlay.style.position = position;
 
             return overlay;
         },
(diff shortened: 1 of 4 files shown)


src/aria/utils/overlay/Overlay.js
```diff
@@ -70,7 +70,16 @@ module.exports = Aria.classDefinition({
...
-            overlay.style.position = params.position || "fixed";
+
+            // absolute positioning is preferred over fixed since Safari Mobile doesn't repaint fixed elements on scroll
+            var position = params.position || "absolute";
+
+            // Place it in the top-left corner of the page, otherwise if it happens to be inserted far in the DOM, it
+            // may temporarily create unnecessary scrollbars on the page - this may affect proper positioning of the
+            // overlay. We'll set left/top later.
+            overlay.style.left = 0;
+            overlay.style.top = 0;
```
divdavem: Rather than setting a temporary position, isn't it possible to append the item to DOM after having positioned it? (changing the order of the calls in the constructor)

jakub-g: In  https://github.com/jakub-g/ariatemplates/commit/67d766837ebbb67b5e7517443248e5b4d5941ced I reversed the order of append+setposition to setposition+append and this made some tests fail
https://travis-ci.org/ariatemplates/ariatemplates/builds/47137563

I'd prefer to not keep the current behavior hence since changing it could break compatibility

jakub-g:
Regarding the suggestion of using `isDevice()` instead of `isIpad || isAndroid`, #1299 says that `isDevice` is not reliable. Let's wait until this gets fixed then ( @ymeine is working on it AFAIR? )

jakub-g:
Landed in master as https://github.com/ariatemplates/ariatemplates/commit/84ffd08ec7f10a8dcaeb6bad6a8375ae95ad60e4 and  https://github.com/ariatemplates/ariatemplates/commit/ccf2df7ae14c3a138e84fc6732459e13154e3e5b

PR #1388, opened 2015-01-12
Fixes #1379 device orientation detection

### LINKED ISSUE #1379: aria.utils.Orientation not reliable
The current implementation of screen orientation detection is not working consistently across devices because:
- it relies on the `orientationchange` event which is Webkit only
- the meaning of values returned in `window.orientation` are device-specific (i.e. 0 means portrait for an iPad but landscape for the Nexus 10.)

A better implementation should rely on the `resize` event and check the window's `innerWidth`/`innerHeight` ratio to distinguish landscape from portrait.

olaf-k:


olaf-k:
Short summary of changes:
- `aria.utils.Orientation` is deprecated
- `aria.utils.Device`'s `isPortrait()` has been updated
- `aria.utils.Device` can raise `orientationchange` event

Orientation is computed using viewport width/height ratio and changes are monitored using `onresize`.  I've also dropped the useless `screenOrientation` property of the original event.

src/aria/utils/Orientation.js
```diff
@@ -33,6 +37,8 @@ module.exports = Aria.classDefinition({
      * Adding a listener while initializing Orientation to listen to the native orientationchange event.
      */
     $constructor : function () {
+        this.$logWarn("This class is deprecated and will be removed from the framework, please use aria.utils.Device instead");
+
```
jakub-g: I'd additionally wrap the whole file (i.e. paste this in the first/last line) in sth like 

`/* BACKWARD-COMPATIBILITY-BEGIN GH-1388 */`
...
`/* BACKWARD-COMPATIBILITY-END GH-1388 */`

we usually do that when the whole class is deprecated

jakub-g: Also I'd change `This class is deprecated` to `aria.utils.Orientation class is deprecated`

olaf-k: - Backward compatibility: sure
- Warning: this is on purpose as $logWarn already indicates the class name
warning

olaf-k:
Landed in 55399ac5e6098a5306928cbe16e863f87eb342df

PR #1389, opened 2015-01-12
fix Wrong error message when the error occurs in a parent template

fbasso:


test/aria/templates/inheritance/logs/LogsTestCase.js
```diff
@@ -0,0 +1,74 @@
...
+                        "The message should be %2 instead of %1");
+
+
+
+                logs.length = 0;
+            }
+            this._end();
+        },
+
+        _end : function () {
```
divdavem: An already existing method called `end` (cf [here](http://ariatemplates.com/aria/guide/apps/apidocs/#aria.jsunit.TemplateTestCase:end:method)) does exactly the same thing as your `_end` method, so you should probably use it and remove this method here.

test/aria/templates/inheritance/logs/ChildTemplate.tpl
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+{Template {
+  $classpath: "test.aria.templates.inheritance.logs.ChildTemplate",
+  $extends: "test.aria.templates.inheritance.logs.ParentTemplate",
```
divdavem: There is an extra comma at the end of this line (which makes IE7 fail!).

test/aria/templates/inheritance/logs/ParentTemplate.tpl
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+{Template {
+    $classpath: "test.aria.templates.inheritance.logs.ParentTemplate",
```
divdavem: There is an extra comma at the end of this line (which makes IE7 fail!).

divdavem:
:ok: cd4cbc73d1b5c3ec97dc4b350f7ea5fe5eff3fdb seems fine for me.

PR #1390, opened 2015-01-14
Dropdown free size was comupted incorrectly in case of vertical scrollbars

flongo:
Method `aria.utils.Size.setContrains` was not correctly taking into account the presence of a vertical scrollbar when the haight of the element is bigger than the maximum one allowed.

### DIFF
--- a/test/aria/widgets/form/autocomplete/popupWidth/AdaptToContentWidthTest.js
+++ b/test/aria/widgets/form/autocomplete/popupWidth/AdaptToContentWidthTest.js
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2013 Amadeus s.a.s.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+Aria.classDefinition({
+    $classpath : "test.aria.widgets.form.autocomplete.popupWidth.AdaptToContentWidthTest",
+    $extends : "aria.jsunit.RobotTestCase",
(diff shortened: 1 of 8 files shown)


test/aria/widgets/form/autocomplete/popupWidth/AdaptToContentWidthTplScript.js
```diff
@@ -0,0 +1,32 @@
...
+    $classpath : "test.aria.widgets.form.autocomplete.popupWidth.AdaptToContentWidthTplScript",
+    $dependencies : ["test.aria.widgets.form.autocomplete.popupWidth.CustomResourceHandler"],
+    $destructor : function () {
+        this.handler.$dispose();
+    },
+    $prototype : {
+
+        getHandler : function () {
+            if (!this.handler) {
+                this.handler = new test.aria.widgets.form.autocomplete.popupWidth.CustomResourceHandler(this.$getId('messages'));
```
divdavem: Is the `this.$getId('messages')` parameter really useful? I don't understand where it is used.

test/aria/widgets/form/autocomplete/popupWidth/AdaptToContentWidthTest.js
```diff
@@ -0,0 +1,86 @@
...
+                        scope : testCase
+                    });
+                }
+            });
+        },
+
+        _afterDropdownOpen : function () {
+            this._checkDropdownWidth();
+            var field = this.getInputField("ac");
+            this.synEvent.execute([["type", field, "[back]a"]], {
```
divdavem: The `[back]` key is supported in the `aria.jsunit.RobotPhantomJS` implementation of the Robot but not in the implementations extending `aria.jsunit.RobotJava`. Maybe you should use `[BACK_SPACE]` instead.

divdavem:
It seems to both work better and simplify the code. It is very good. Thank you for this PR.

flongo:
Thank you for your comments @divdavem. I have applied the suggested changes, I will integrate as soon as the Travis builds terminates successfully

divdavem:
It seems that changes introduced in commit 43add31528fe5e2baccf6471d7133356220bf1a4 make the `test.aria.widgets.container.dialog.resize.test1.DialogOnResizeTestCase` test fail on all browsers (apart from PhantomJS).

For example, with Chrome:
```
Assert #6 failed : offsetLeft should be around 50 instead of 41
Assert #7 failed : offsetWidth should be around 300 instead of 318
```

PR #1391, opened 2015-01-14
fix #1391 run Phantom-unfriendly tests also in Travis

jakub-g:
Some tests were never run locally due to some issues with PhantomJS.

This PR makes `npm test` run those tests with Chrome and Firefox, via Selenium Java Robot.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,20 @@ language: node_js
 node_js:
   - 0.10
 before_install:
+# Changing the /etc/hosts file is needed for the selenium Firefox driver to work correctly
+# (see https://code.google.com/p/selenium/issues/detail?id=3280)
+  - (echo "127.0.0.1 localhost" && cat /etc/hosts) > /tmp/hosts && sudo mv /tmp/hosts /etc/hosts
+# Settings needed for graphic output in Chrome/Firefox
+  - export DISPLAY=:99.0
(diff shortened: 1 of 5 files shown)


divdavem:
I think it would not be very difficult to also make those tests run on travis.
We could either use Firefox (as we do as part of the attester test suite [cf .travis.yml here](https://github.com/attester/attester/blob/master/.travis.yml)) or Chromium (which could probably be installed with `sudo apt-get install chromium-browser` in `before_install`).

attester-chrome.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+if [ "$TRAVIS" = "true" ]; then
+    exit
+fi
+
+(node node_modules/attester/bin/attester.js test/attester-phantom-complementary.yml --env package.json &) && selenium-java-robot --browser 'chrome'
```
divdavem: Note that you can use the `--robot-browser` parameter of attester instead of invoking yourself the `selenium-java-robot` command.

divdavem: Maybe the full command line could be stored directly in the `package.json` file (especially if the check of the `$TRAVIS` variable is no longer required if we also run those tests on travis).

divdavem:
:ok: 8b47d181994b6c23c433e9026199b5e526bee402 seems fine for me

jakub-g:
@divdavem thank you for the code review remarks

Regarding Travis, it has Chromium and Firefox ESR by default so there's no need to install anything in `.travis.yml`

PR #1392, opened 2015-01-22
Dialog widget: clicking on the close button interferes with drag/drop

divdavem:
This PR fixes the following issue: on some browsers, clicking on the close button of the dialog is randomly interpreted as the beginning of a drag/drop operation, preventing the dialog from being closed.


fbasso:
00bf155 fine for me

PR #1393, opened 2015-01-22
Data not passed to template from placeholder page engine configuration

divdavem:
The `data` property, specified in [aria.pageEngine.CfgBeans.Placeholder](http://www.ariatemplates.com/aria/guide/apps/apidocs/#aria.pageEngine.CfgBeans:Placeholder), was not taken into account by the page engine. This pull request fixes the issue, the `data` property is now properly passed to the template.


fbasso:
b8fe527 seems fine to me

PR #1394, opened 2015-01-26
fix #1381 User input discarded in AutoComplete without freeText when clicking outside

### LINKED ISSUE #1381: User input discarded in AutoComplete without freeText when clicking outside
* Go to: [this sample](http://snippets.ariatemplates.com/samples/github.com/ariatemplates/documentation-code/tree/master/samples/widgets/autocomplete/preselectAutofill)
* Unset the "freeText" option
* Type something which corresponds to the beginning of a suggestion, for example: "slo"
* Leave the field by clicking outside

Expected behavior: the content of the field is kept and the field is in error, as it does not match a suggestion (this is the behavior when leaving the field with the TAB key instead of clicking outside)

Actual behavior: the content of the field is discarded, which is bad for the user experience, as the user is simply loosing what he has just typed

fbasso:


divdavem:
:ok: 774108c40c23084c98c72c98fcf57782b63d1e9c seems fine for me

fbasso:
landed to aee5efd

PR #1396, opened 2015-01-29
Changing error message when a syntax error happens in a template

divdavem:
The previous error message (stating that the module did not contain the expected class) was confusing. It is now clearer.


fbasso:
5127fa7 fine for me

PR #1399, opened 2015-02-02
fix failing test DialogOnResizeTestCase

lsimone:
The test was failing after recent changes on `aria.utils.Size` ( 43add31528fe5e2baccf6471d7133356220bf1a4 ) that take into account the scrollbar to compute the size of elements.
It has been fixed (and refactored).


divdavem:
:ok: 08c21564cdc21653bbc44c7ffb59492c7a4b6f34 seems fine for me

lsimone:
This landed in master with commit: 236d766adfe76ae713ef603910079cd3e2caea6e

PR #1400, opened 2015-02-02
refactor Splitting test.aria.widgets.form.autocomplete.errorhandling.AutoComplete

divdavem:
A new scenario was appended to this test in commit aee5efd724c2539b4b53a9725a2af048f8b18c5d, but this scenario seems to fail on IE. Therefore, this commit separates into 2 tests the old and the new scenario so that it is clearer to understand what is failing, and if there is a regression.


lsimone:
it seems :ok: to me

PR #1404, opened 2015-02-03
Fix test.aria.widgets.container.checkContent.DivTest on Safari

divdavem:
Fixes test.aria.widgets.container.checkContent.DivTest on Safari (in which scroll bars take no space).


benouat:
@divdavem Seems :ok: for me. I would still make sure it is working fine on Chrome and FF OSX.

PR #1405, opened 2015-02-03
Fix #1388 for old IE

olaf-k:
Viewport dimensions were not reported properly for IE7/8.

### DIFF
--- a/src/aria/utils/Device.js
+++ b/src/aria/utils/Device.js
@@ -15,6 +15,7 @@
 var Aria = require("../Aria");
 var ariaCoreBrowser = require("../core/Browser");
 var ariaUtilsEvent = require("./Event");
+var ariaUtilsDom = require("./Dom");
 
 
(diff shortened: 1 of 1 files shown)


benouat:
@olaf-k fine for me. you can merge it :ok:

src/aria/utils/Device.js
```diff
@@ -246,7 +247,10 @@ module.exports = Aria.classDefinition({
          * @return {Boolean}
          */
         isPortrait : function (forceCheck) {
-            if (forceCheck) return Aria.$window.innerHeight > Aria.$window.innerWidth;
+            if (this._isPortrait === null || forceCheck) {
```
ymeine: Better check for `null` like this: `this._isPortrait == null`.

lsimone:
this landed with commit 92ab46a3b1c93ed89103427ab56d55bd846896d5

PR #1407, opened 2015-02-03
fix SplitterTestMoveLeftRight and SplitterTestMoveUpDown

divdavem:
When SplitterTestMoveLeftRight and SplitterTestMoveUpDown are run in attester, probably because of the added iframe, the actual splitter seems to be a few pixels on the right or on the bottom of the mouse position, resulting in a test failure. With this commit, instead of starting the drag operation at the edge of the handle, the mouse is now centered on the handle which makes those tests more reliable.


fbasso:
2774d27 fine for me

PR #1408, opened 2015-02-04
Allowing the robot not to be available immediately (waits for 2s)

divdavem:
Sometimes, if the first test to be executed by a browser is a robot test, it can fail with the error "The robot is not available." even if the browser was started with `selenium-java-robot` because `selenium-java-robot` has not yet injected the `SeleniumJavaRobot` object in the page. This PR makes sure we wait up to 2s before raising the error that the robot is not available.


fbasso:
ab10210 fine for me

PR #1411, opened 2015-02-13
fix #1411 load extrascripts in Classic Test Runner

jakub-g:
This PR adds the possibility to load in the Classic Test Runner
the extrascripts exposed in attester's config, when CTR is run
in isolated mode.

Previously, the extrascripts were injected by attester only to CTR's
index page, but CTR had no possibility to propagate those scripts
into the iframe when tests were run with `#runIsolated=true` in the URL.

This commit needs https://github.com/attester/attester/pull/121
integrated in attester in order for it to properly expose `before`
and `after` extrascripts.

---

The other 2 commits are just minor visual improvements

### DIFF
--- a/src/aria/jsunit/TestWrapper.js
+++ b/src/aria/jsunit/TestWrapper.js
@@ -16,6 +16,7 @@ var Aria = require("../Aria");
 var ariaUtilsFrameATLoader = require("../utils/FrameATLoader");
 var ariaJsunitAssert = require("./Assert");
 var ariaCoreTimer = require("../core/Timer");
+var ariaUtilsScriptLoader = require("../utils/ScriptLoader");
 
 
 /**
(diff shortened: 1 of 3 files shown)


jakub-g:
@divdavem I added a new commit removing `force` and just not storing scripts in array if they don't come from the main document. What do you think about it now?

It's a bit artificial to do it IMO but at least it doesn't break compat.

src/aria/jsunit/TestWrapper.js
```diff
@@ -51,6 +62,41 @@ module.exports = Aria.classDefinition({
...
+         * and injects them into an iframe. If there are no such scripts, it does nothing.
+         * It calls the callback
+         * @param {Document} subDocument document element of the iframe
+         * @param {String} whereToInsert expected value of `data-extrascript` attribute
+         * @param {aria.core.CfgBeans.Callback} callback
+         */
+        _injectExtraScriptsIntoFrame : function (subWindow, whereToInsert, callback) {
+            var document = Aria.$window.document;
+            var subDocument = subWindow.document;
+            var scripts = document.querySelectorAll('script') || document.scripts;
```
divdavem: `querySelectorAll` is not supported by IE7, whereas `document.scripts` is not supported by Firefox < 9.
See https://developer.mozilla.org/en-US/docs/Web/API/Document/scripts and https://developer.mozilla.org/en/docs/Web/API/Document/querySelectorAll

I would reverse the order so that the line work in all browsers (including IE7):
```js
var scripts = document.scripts || document.querySelectorAll('script');
```

divdavem: Also, rather than using `querySelectorAll` which is generic for all CSS selectors, you should probably use `getElementsByTagName`.

jakub-g: Good catch!

jakub-g:
@divdavem 
I fixed the `querySelectorAll` issue & rebased (I also reorder the commits)
Build failure is due to MapManagerTest random failure

divdavem:
@jakub-g Thank you.
:ok: 5a5a780f442634463a4435e5b1ac6122a8390e8a (and its 4 parent commits) is fine for me.

jakub-g:
@divdavem thank you for the review!

PR #1414, opened 2015-02-17
fix #1413 Refreshing disposed templates

### LINKED ISSUE #1413: Null config in template refresh
There is an issue when stopping the refresh manager, calling a refresh, then disposing of one of the templates that has been queued for a refresh, then starting the refresh manager again.  

In this scenario the refresh should be simply ignored for the disposed template, however currently this isn't the case and instead an error is thrown because the config object is null.

simonarbuckle:


divdavem:
:ok: 6d6c49e77c1d4b5f7a765e54a5b6cea98ce11eac seems fine for me.

PR #1415, opened 2015-02-20
Package Aria Templates with the ATFillCache builder

divdavem:
This pull request changes the format in which Aria Templates is packaged. Instead of using the previous ATMultipart format (which does not allow packages to be evaluated in a script tag), it is now using the ATFillCache format (cf [this commit in atpackager](https://github.com/ariatemplates/atpackager/commit/9fe83ae8fb9cc0630cc4dee3617139bccb8f5b4c)) which fills the cache when packages are directly evaluated in a script tag.


lsimone:
LGTM :ok:

PR #1416, opened 2015-02-23
Failing TemplateManagerTest and SectionAsContainerTest on IE 7

divdavem:
This PR fixes a regression introduced in commit f9a167054598fdab57137f9a06cec83d36b6477b for TemplateManagerTest and SectionAsContainerTest on IE 7.

On IE 7, the toString method of an error object returns no information about the error (only the string "[object Error]"). It is necessary to use the description property instead.

This PR depends on the following pull request on noder-js: https://github.com/ariatemplates/noder-js/pull/30


simonarbuckle:
:ok: 94621102ce5dca60419fd15e0c02bd1bb8c81c97 seems fine for me.

PR #1419, opened 2015-02-25
Allowing compiled templates to use the new syntax (with require)

divdavem:
This pull request makes sure it is possible to load already compiled templates converted to the new syntax (with require).

### DIFF
--- a/src/aria/Aria.js
+++ b/src/aria/Aria.js
@@ -821,6 +821,24 @@ var Aria = module.exports = global.Aria;
         return res;
     };
 
+    /**
+     * This function is useful when either a classpath or a reference to a class is expected, to convert such a value to
+     * a classpath. If the parameter is a reference to an Aria Templates class or interface, its classpath is returned.
+     * Otherwise, the parameter is returned as is.
+     * @param {String|Object|Function} classpathOrCstr Classpath or reference to a class.
+     * @return {String} Classpath corresponding to classpathOrCstr
+     */
+    Aria.getClasspath = function (classpathOrCstr) {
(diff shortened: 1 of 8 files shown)


src/aria/Aria.js
```diff
@@ -821,6 +821,21 @@ var Aria = module.exports = global.Aria;
         return res;
     };
 
+    /**
+     * Returns the classpath of the given reference, if it is not already a classpath.
+     * @param {String|Object} classpathOrCstr Classpath or reference to a class.
+     */
+    Aria.getClasspath = function (classpathOrCstr) {
```
ymeine: Misses a comment about the case where the input wouldn't be a string and the classpath couldn't be detected. Depending on the desired behavior, according code might miss as well.

src/aria/core/loaders/GeneralTplPreprocessor.js
```diff
@@ -15,7 +15,7 @@
 
 var asyncRequire = require("noder-js/asyncRequire").create(module);
 var firstComment = /^\s*\/\*[\s\S]*?\*\//;
-var alreadyGeneratedRegExp = /^\s*Aria\.classDefinition\(/;
+var alreadyGeneratedRegExp = /^\s*(var\s+|Aria\.classDefinition\()/;
```
ymeine: Even if not changing the behavior in current code state, it is better to clearly define whether the content between parentheses should be a capturing group or not, which means by extension: __do we intend to be able to fetch the corresponding matched content__. If yes, keep it as is. Otherwise, use a non-capturing group (`(?:)`).

ymeine:
Good for me!

PR #1420, opened 2015-03-03
fix The bound list of a simple select takes only every other item on IE <= 9

fbasso:


divdavem:
:ok: de7921e1332a7b51e38f85f6a536cbf7c344dec8 seems fine for me

PR #1421, opened 2015-03-03
feat Refactoring of Browser and Device classes: deletion of deprecated Orientation class & removal of deprecated code usage

ymeine:
There is another commit dealing with the deletion of deprecated code, but this one completes it by removing some code which has been missed.

The compatibility commit can be reverted as soon as the deprecation period has ended. It completes commit 608c176 and the latter should be considered regarding the date of the deprecation.

In addition to that, there was some residual usage of deprecated code inside our own class, so this has been removed as well (in practice this was the `environment` property of the class `Browser`).

__Finally__ this means that only the two following commits will have to be reverted (instead of initially planned one 7369fa9eb35c5e6f64a2737a608320b71ff12323): 

- da1a907e3bf371619e2fa758f3fb97eca4707b43
- 3ee6d19a23aaaee85ecd55e8dd90aead0bcaf556


lsimone:
7dc4bab53528e26da72c6b2174ceff511704d543 looks good to me :+1:

PR #1423, opened 2015-03-06
Allow CSSMgr to reuse existing tags

flongo:
It is important for an application to decide the order
in which widgets and templates-related css rules
(coming from CSS templates) are inserted in the head tag.

In order to do that, the application can insert tags with the correct id. This commit allows to use existing tags for the css insertion.

### DIFF
--- a/src/aria/templates/CSSMgr.js
+++ b/src/aria/templates/CSSMgr.js
@@ -18,6 +18,7 @@ var ariaUtilsArray = require("../utils/Array");
 var ariaUtilsObject = require("../utils/Object");
 var ariaUtilsAriaWindow = require("../utils/AriaWindow");
 var ariaCoreClassMgr = require("../core/ClassMgr");
+var ariaUtilsDom = require("../utils/Dom");
 
 /**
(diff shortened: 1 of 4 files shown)


src/aria/templates/CSSMgr.js
```diff
@@ -660,16 +664,16 @@ module.exports = Aria.classDefinition({
...
                 tag.type = "text/css";
                 tag.media = "all"; // needed as the default media is screen in FF but all in IE
 
                 head.appendChild(tag);
                 tag = head.lastChild;
-
-                this.__styleTagPool[tagName] = tag;
             }
 
+            this.__styleTagPool[tagName] = tag;
```
flongo: Should we make sure that the tag has the correct `type` and `media`?

divdavem:
:ok: c9c690e8f734e88e7f0b621fd0135d299685f657 seems fine for me.

src/aria/templates/CSSMgr.js
divdavem: Well, this should probably be done by the code which creates the tag.

divdavem:
:ok: abd402ae9ca23cfc1e9b18ed64461c666f145246 seems fine for me

PR #1424, opened 2015-03-09
Configuration beans link in documentation

mathbruyen:
Few configuration beans were incorrectly referred to by the API documentation (using a dot instead of a colon). Fixed for beans called CfgBeans.


lsimone:
it looks good to me, it can be integrated :ok:

PR #1425, opened 2015-03-10
feat # validator messages at application level

lsimone:
This commit provides the possibility to globally set messages for validators.
Validators will show the message with highest priority in this order:
1) Local message
2) Global message
3) Hardcoded default

close #


simonarbuckle:
:ok: 2f1abda2e2a0f9ebf566a374dc3fa4bead43ded4 seems fine for me

PR #1427, opened 2015-03-12
feat Customization of widget's error messages

ymeine:
It is now possible for a user to customize the error messages used by the widgets.

There are 3 levels of error messages specifications, the two firsts being user configuration while the last is the hardcoded default.

This is resolved in this order of precedence:

- local: widget's configuration
- global: application's environment
- hardcoded: widget's resources

### DIFF
--- a/src/aria/widgets/controllers/TextDataController.js
+++ b/src/aria/widgets/controllers/TextDataController.js
@@ -15,6 +15,10 @@
 var Aria = require("../../Aria");
 var ariaWidgetsControllersReportsControllerReport = require("./reports/ControllerReport");
 var ariaUtilsType = require("../../utils/Type");
+var ariaWidgetsWidgetsRes = require("../../$resources").file(__dirname, "../WidgetsRes");
+var ariaWidgetsSettings = require("../environment/WidgetSettings");
+
+
 
 
 /**
@@ -22,9 +26,12 @@ var ariaUtilsType = require("../../utils/Type");
  */
 module.exports = Aria.classDefinition({
     $classpath : "aria.widgets.controllers.TextDataController",
+    $resources : {
+        res : ariaWidgetsWidgetsRes
+    },
     $events : {
         "onCheck" : {
-            description : "Notifies that controller has finished am asynchronouse check (internal, of the value or of a keystroke)",
+            description : "Notifies that controller has finished an asynchronous check (internal, of the value or of a keystroke)",
             properties : {
                 report : "{aria.widgets.controllers.reports.ControllerReport} a check report"
             }
@@ -39,6 +46,8 @@ module.exports = Aria.classDefinition({
             value : null,
             displayText : ''
         };
+
+        this.setDefaultErrorMessages();
     },
     $prototype : {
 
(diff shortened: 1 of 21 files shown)


src/aria/widgets/controllers/TextDataController.js
```diff
@@ -222,6 +256,104 @@ module.exports = Aria.classDefinition({
                 report : report,
                 arg : arg
             });
+        },
+
+        /**
+         * Retrieves the requested error message from different sets of configuration ordered by precedence.
+         *
+         * <p>
+         * An error message value is requested from an this error message name.
```
lsimone: maybe "an error message name"?

src/aria/widgets/controllers/TextDataController.js
```diff
@@ -222,6 +256,104 @@ module.exports = Aria.classDefinition({
...
+         *
+         * <p>
+         * Note that since there are hardcoded values, there will always be an error message for a valid message name. However, if the message name is not supported, an <em>unefined</em> value is returned in the end.
+         * </p>
+         *
+         * @param {String} errorMessageName The name of the error message to retrieve (it is the key to the requested message in the collection).
+         *
+         * @return {String} The retrieved error message if any, <em>undefined</em> if the message name is not supported.
+         */
+        getErrorMessage : function (errorMessageName) {
```
lsimone: This function could be generalized and moved in a utils class, since the configuration and logical parts are well decoupled.

test/aria/widgets/form/BaseDefaultErrorMessagesTest.js
```diff
@@ -0,0 +1,145 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+Aria.classDefinition({
+    $classpath : "test.aria.widgets.form.BaseDefaultErrorMessagesTest",
+    $extends : "aria.jsunit.TemplateTestCase",
+    $dependencies : ["aria.utils.Array", "aria.utils.Json"],
+    $constructor : function (widgetName, messagesList) {
```
lsimone: I would add just 2 lines to briefly explain what are supposed to do new tests that extend this one (or modify the ones already implemented) and what they should define.

lsimone:
apart from the small possible improvements I suggested, d9d8903ed18114371c612d58f1559a802c91c85b looks good to me :+1:

test/aria/widgets/form/BaseDefaultErrorMessagesTest.js
ymeine: Described the role of the two parameters, I think this is enough in this case.

src/aria/widgets/controllers/TextDataController.js
ymeine: "this" actually, since it refers to the error massage mentioned at the beginning of the sentence.

src/aria/widgets/controllers/TextDataController.js
ymeine: Would be great (with a little bit of generalization though), as for many other parts of our code! :) But we'll leave for the future... maybe...

PR #1428, opened 2015-03-19
feat Improved the widget's default error messages customization and added backward compatibility

ymeine:
## Improvement

The implementation code is now cleaner and more documented, as well as for the test. A little bit of redundancy checking has been introduced in the tests just to make sure that if the order of some methods changes or if some are removed or whatever the tests will still work and be comprehensive.

Also fixed a small mistake in AutoComplete's default error messages test.

An unwanted `return` statement was remaining at the end of the `$constructor` of the test. It was not breaking anything regarding our class system implementation, but this remains incorrect.

## Backward compatibility

Added backward compatibility for widget's default error messages customization. The non-backward change concerns the keys used inside a resource file to define the hardcoded fallback for default values.

There is no deprecation message or so on because the broken feature was not a public feature, but a commonly agreed hack.

### DIFF
--- a/src/aria/widgets/WidgetsRes.js
+++ b/src/aria/widgets/WidgetsRes.js
@@ -36,6 +36,17 @@ Aria.resourcesDefinition({
             "AutoComplete" : {
                 "validation" : "There is no suggestion available for the given entry."
             }
+            /* BACKWARD-COMPATIBILITY-BEGIN (GitHub #1428) */
+            ,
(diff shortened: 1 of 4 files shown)


divdavem:
:ok: 82250a1d97988a40ebc95155269d17556893f510 and c2438b7c57805f2b9f95b905ccaefed49ce5b91b are fine for me.

PR #1430, opened 2015-03-24
fix Selection was lost when a multiselect was open with the arrow down key

fbasso:
It happens when some checkboxes are checked first : reopen the multiselected with the arrow down key showed all checkboxes unchecked.


divdavem:
:ok: 382a7defdeb2480206ae892ba982e4c59f76aa3c seems fine for me.

PR #1432, opened 2015-03-26
fix #1412 mouse events are now attached to the window directly

### LINKED ISSUE #1412: aria.utils.dragdrop.Drag doesn't detect move/dragend outside of window
When using a aria.utils.dragdrop.Drag, if the user starts dragging and releases the mouse button outside of the browser window, the dragEnd event is not fired.

Also if you mousemove outside of the viewport, the move event is not triggered.

In order to detect this properly, the mouseup and mousemove events in aria.utils.Mouse should be attached to window instead of window.document.body.

simonarbuckle:


lsimone:
1381bb840321604b95587fb1c70be6ffbbd294cc looks good to me :+1:

PR #1437, opened 2015-04-01
fix #1436 Check widget destruction before reacting to data change

### LINKED ISSUE #1436: Error when RadioButton click triggers container section refresh
When a `html:RadioButton` widget is bound to a certain part of the data model and the section that contains it has its refresh bound to the same data model, there is an error in the console: the data model listeners are called after the widget destruction.

Here is a sample: http://instant.ariatemplates.com/anonymous/abb14c5c1dcd523959ea/5c966c

flongo:
This commit introduces a check on the status of the widget before reacting to data change.

It can happen that the `_notifyDataChange` method is called after the widget destruction.

### DIFF
--- a/test/aria/html/radioButton/listenerAfterDestruction/ListenerCalledAfterDestruction.tpl
+++ b/test/aria/html/radioButton/listenerAfterDestruction/ListenerCalledAfterDestruction.tpl
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2012 Amadeus s.a.s.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


test/aria/html/radioButton/listenerAfterDestruction/ListenerCalledAfterDestruction.tpl
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2012 Amadeus s.a.s.
```
lsimone: copyright 2015 here

test/aria/html/radioButton/listenerAfterDestruction/ListenerCalledAfterDestructionTest.js
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2012 Amadeus s.a.s.
```
lsimone: same thing here: copyright 2015

lsimone:
5d642b2948b7e0651775c4128a99165a93e09d4c looks good to me :+1:

PR #1446, opened 2015-04-07
Disabled state in html:RadioButton widget

flongo:
When two `html:RadioButton`s disabled state is bound to the same non-initialized data model, an error was triggered.
This was due to the fact that when the `initWidget` method of the first one was called, a `setValue` was done on the data model, which triggered some listeners to modify the disabled state of the second one, for which the `initWidget` method had not been called yet, so that property `this._domElt` had not been set yet.


divdavem:
:ok: 8bf4ab654a856a77bdfd9c384d9e02f29d2d771b seems fine for me

PR #1448, opened 2015-04-13
Changing the template compilation result to use the new syntax

divdavem:
This pull request changes template class generators to generate code with the new syntax (with `require`) instead of the old one (with `$dependencies`).

A specific use case (tested by `test.aria.templates.reloadResources.ReloadResourcesTestCase`) requires some changes in noder-js (cf ariatemplates/noder-js#31). It is handled in a separate commit.

Before integrating this PR:
- [x] wait for ariatemplates/noder-js#31 to be integrated in noder-js
- [x] use the new version of noder-js

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "at-noder-converter": "1.0.1",
     "atpackager": "0.2.6",
     "gzip-js": "0.3.2",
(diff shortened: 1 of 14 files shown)


fbasso:
Fine for me.

PR #1449, opened 2015-04-17
feat Improvement for the data utility, now only validates the model once each time validateModel is called

simonarbuckle:
This new feature adds a map to the data utility which stores data model objects that have been validated.  The map is used to check and enforce a model object does not get validated twice.  This resolves the following:

- recursive references within a data model object leading to infinite loops and RangeErrors
- a data model object referencing a previously validated data model object

**Illustration**:

          * (root)
        /   \
       A     B (reference to A)  
      /   
     C (reference to *)

### DIFF
--- a/src/aria/utils/Data.js
+++ b/src/aria/utils/Data.js
@@ -18,7 +18,7 @@ var ariaUtilsPath = require("./Path");
 var ariaUtilsJson = require("./Json");
 var ariaCoreJsObject = require("../core/JsObject");
 var ariaUtilsArray = require("./Array");
-
+var ariaUtilsStackHashMap = require("./StackHashMap");
 
 /**
  * Handles the link between the validators and the data model.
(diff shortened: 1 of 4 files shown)


src/aria/utils/Data.js
```diff
@@ -366,24 +366,45 @@ module.exports = Aria.classDefinition({
          * @param {Object} messages
          * @param {Array} groups an optional parameter that contains the group(s) to be validated, if no group is
          * specified then validation will occur on the model as normal.
-         * @param {Boolean} stopOnError an optional parameter that (when set to true) stops the validation as soon as an error is detected in
-         * the model provided (defaults to false)
+         * @param {Boolean} stopOnError an optional parameter that (when set to true) stops the validation as soon as an
+         * error is detected in the model provided (defaults to false)
+         * @param {Object} Map stores a reference to an already validated part of the model
```
divdavem: There is no map parameter on this function, so the jsdoc should not mention it.

src/aria/utils/Data.js
```diff
@@ -366,24 +366,45 @@ module.exports = Aria.classDefinition({
...
+        /**
+         * validates an object within the data model, including all fields values and validation stacks for those fields
+         * @param {Object} dataHolder
+         * @param {Object} messages
+         * @param {Array} groups an optional parameter that contains the group(s) to be validated, if no group is
+         * specified then validation will occur on the model as normal.
+         * @param {Boolean} stopOnError an optional parameter that (when set to true) stops the validation as soon as an
+         * error is detected in the model provided (defaults to false)
+         * @param {Object} Map stores a reference to an already validated part of the model
+         * @protected
```
divdavem: There is no need to have the `@protected` keyword here, as this is automatically deduced from the underscore in the name of the method.
The type of the map parameter is `aria.utils.StackHashMap`.

src/aria/utils/Data.js
```diff
@@ -396,20 +417,20 @@ module.exports = Aria.classDefinition({
          * @param {String} parameter
          * @param {Object} messages
          * @param {Array} group an optional parameter that contains the group(s) to be validated.
-         * @param {Boolean} stopOnError an optional parameter that (when set to true) stops the validation as soon as an error is detected in
-         * the model provided (defaults to false)
+         * @param {Boolean} stopOnError an optional parameter that (when set to true) stops the validation as soon as an
+         * error is detected in the model provided (defaults to false)
```
divdavem: The map parameter is not mentioned in the jsdoc here.

divdavem:
:ok: f19fd858ac7c73779fbc2aa44f5d2d78ba579c03 seems fine for me

PR #1451, opened 2015-04-20
Avoid useless recursion in aria.utils.Json.inject if objects are identical

flongo:
If you inject an object into another setting the merge flag to true, if the objects are identical and contain cyclic references, then the maximum call stack size is exceeded.

### DIFF
--- a/src/aria/utils/Json.js
+++ b/src/aria/utils/Json.js
@@ -18,7 +18,6 @@ var ariaUtilsType = require("./Type");
 var ariaUtilsArray = require("./Array");
 var ariaUtilsObject = require("./Object");
 
-
 (function () {
 
     // shortcut to array & type utils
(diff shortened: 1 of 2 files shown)


divdavem:
:ok: ff63f58552d4c26a45a4ae2e55e93ceea7d0a567 seems fine for me

divdavem:
@flongo Thank you for your pull request. It solves the issue in case the source and destination are identical.
However, in the general case, `aria.utils.Json.inject` does not currently support cyclic sources if the destination is also cyclic in the similar way (without identical objects).

For example, the following piece of code still has the infinite recursion problem:

```js
function f() {
   var a = {
      b: "c"
   }
   a.d = a;
   return {
      a : a
   };
};
aria.utils.Json.inject(f(), f(), true);
```

If such a use case needs to be supported as well, maybe, we could store visited objects with an instance of `aria.utils.StackHashMap` in a way similar to what is done in b384881ce9b64d95f2afd7aa5a21de48f5519b26...

flongo:
Thank you @divdavem for the remark. For the time being we are happy with the identical objects fix. Can I merge?

divdavem:
@flongo yes, you can merge

PR #1452, opened 2015-04-21
feat #1452 Classic Test Runner: explicitly declare wlibs

jakub-g:
Explicitly declare dependencies so that an application using the runner does not need to depend on AriaLib in order to run the tester

See https://github.com/attester/attester/pull/124

I tested the fix on our application;
When calling  in our TestBootstrap.js `aria.core.AppEnvironment.updateEnvironment({   defaultWidgetLibs: { ...`  without AriaLib it fails with TEMPLATE ERROR when loading Classic Test Runner. With this commit, the test runner loads fine even without AriaLib added on our side.

### DIFF
--- a/src/aria/tester/runner/view/config/Config.tpl
+++ b/src/aria/tester/runner/view/config/Config.tpl
@@ -16,6 +16,7 @@
 // TODOC
 {Template {
     $classpath:'aria.tester.runner.view.config.Config',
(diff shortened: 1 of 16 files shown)


divdavem:
:ok: 7dd9941de448000c63c7c00190f5e7ee6efc17c9 seems fine for me.

PR #1453, opened 2015-04-22
fix Improve performance in aria.utils.IdManager

fbasso:


test/aria/utils/IdMgr.js
```diff
@@ -0,0 +1,92 @@
...
+    $extends : "aria.jsunit.TestCase",
+    $dependencies : ["aria.utils.IdMgr", "aria.utils.Array"],
+    $prototype : {
+
+        /**
+         * Test IdManager with simple complete use case
+         */
+        test_getId : function () {
+
+            var nbToCreate = 50;
```
lsimone: maybe we could declare the var name in upper case, since it is a constant?

test/aria/utils/IdMgr.js
```diff
@@ -0,0 +1,92 @@
...
+            for (i = 0; i < nbToCreate; i++) {
+                contains = contains || aria.utils.Array.contains(pool2, pool1[i]);
+            }
+            this.assertFalse(contains, "Prefixing failed.");
+
+            // release and recreate
+            for (i = 0; i < nbToCreate; i++) {
+                idMgr1.releaseId(pool1[i]);
+                /*
+                // release twice, should not fail
```
lsimone: why commented out?
is releasing twice handled?

lsimone:
except for the comments I wrote, it looks good to me :+1:

PR #1454, opened 2015-04-23
feat atLoader injector script to load AT packages

lsimone:
atLoader is used to synchronously load Aria Templates classes that are not included
in the AT package.
It is sufficient to insert the array of classes as the innerHTML of the tag
used to load this script.
All the classes' dependencies have to be added to this array.
No comments or other sintax is allowed in it except for the array itself.


divdavem:
:ok: e9b2e6a04da22a9ae7dc5fc581f70507a8679628 seems fine. You can integrate it if the travis build is ok.

PR #1456, opened 2015-04-24
Fixing Drag class dispose process

divdavem:
This commit fixes the following error:
```
Unable to get property 'replace' of undefined or null reference
```
which happened when running the following test: `test.aria.widgets.container.dialog.resize.test5.OverlayOnResizeScrollTestCase`


lsimone:
LGTM :+1:

PR #1458, opened 2015-04-27
aria.utils.Dom.calculatePosition extra pixels in IE 7

divdavem:
This commit improves the detection of the extra pixels returned by `getBoundingClientRect` in IE 7.

cf http://chunghe.blogspot.fr/2008/03/15-use-getboundingclientrect-if.html
and https://github.com/jquery/jquery/blob/master/src/offset.js#L101-106


lsimone:
LGTM :+1:

PR #1466, opened 2015-04-30
Repositions the popup of the multiselect when clicking on an option

divdavem:
Clicking on a multiselect option often triggers a refresh, which may lead to the multiselect being visually moved. This commit makes sure that, in that case, the popup moves along with the multiselect.


lsimone:
Looks good to me :+1:

PR #1467, opened 2015-05-04
Empty multiselect focus issue

divdavem:
This pull request fixes the following issue: when a `@aria:MultiSelect` widget is empty (i.e. when its `items` property is an empty array) and when clicking on the dropdown icon to open the popup, the widget gets the focused style and keeps it forever (even when the focus is somewhere else).


ymeine:
This is fine for me, you can merge when you are sure the tests pass on all the platforms!

PR #1468, opened 2015-05-04
fix #1468 Prefix URLs in CSS templates feature not working

jakub-g:
Feature implemented in #972 did not work due to invalid CSS
generated: a space between `url` keyword and the parentheses
made the whole CSS rule invalid and hence not applied

----------

As amazing as it might be, it seems that this feature in fact never really worked, or somehow the browsers changed how they interpret `url ('...')` strings in CSS which is quite hard to believe (but never say never)

Consider:
http://output.jsbin.com/meqafekasa/

It has the following HTML:
```
 <div style="width:150px; height: 30px; background-color:green ; background-image: url('http://www.amadeus.com/web/static/82b6bd11/img/amadeus.png')"> </div>
  
 <div style="width:150px; height: 30px; background-color: red;   background-image: url ('http://www.amadeus.com/web/static/82b6bd11/img/amadeus.png')"> </div>
```

The first one does not have space between `url` and left parenthesis, the second one has a space. Tested across Chrome, Firefox and IE9, first is always working and the second never.

### DIFF
--- a/test/aria/templates/css/imgprefix/ImgPrefixTemplateTestCase.js
+++ b/test/aria/templates/css/imgprefix/ImgPrefixTemplateTestCase.js
@@ -25,7 +25,8 @@ Aria.classDefinition({
             template : "test.aria.templates.css.imgprefix.SimpleTemplate"
         });
 
-        this.imgPaths = ["images/bg.jpg", "images/bg.jpg", "images/bg.jpg", "images)/bg.jpg", "images/?bg.jpg", "images'/bg.'jpg", "images\"/bg.\"jpg"];
+        this.imgPaths = ["images/bg.jpg", "images/bg.jpg", "images/bg.jpg", "images)/bg.jpg", "images/?bg.jpg",
+                "images'/bg.'jpg", "images\"/bg.\"jpg"];
 
         // set the app environment
         aria.core.AppEnvironment.setEnvironment({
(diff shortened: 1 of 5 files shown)


jakub-g:
added a test, it is green in chrome, ff, ie9, phantom

leaving up to the team to test in the other browsers

test/aria/templates/css/imgprefix/ImgPrefixTemplateTestCase.js
```diff
@@ -75,12 +79,42 @@ Aria.classDefinition({
...
+
+        _loadThirdTpl : function () {
+            this._replaceTestTemplate({
+                template : "test.aria.templates.css.imgprefix.TemplateWithRealImage"
+            }, this._afterThirdTemplateLoaded);
+        },
+
+        _afterThirdTemplateLoaded : function () {
+            var container = Aria.$window.document.getElementById("container-real-image");
+            var style = aria.utils.Dom.getStyle(container, "background-image");
```
divdavem: `aria.utils.Dom.getStyle` is expecting the property in camel case:

```
var style = aria.utils.Dom.getStyle(container, "backgroundImage");
```

Otherwise, the test fails in Firefox 3 and Firefox 11...

divdavem:
:ok: 203fe01d610d4afd2c558d8a49fd16fa761c4e92 seems fine for me. Thank you!

PR #1471, opened 2015-05-27
fix aria.utils.Date.format didn't work well for the day of the week, when used with the last parameter to true

fbasso:
aria.utils.Date.format(new Date("2015/05/26 23:15:00 GMT+0000"),"EEE d MMM",true) returned `Wed 26 May` instead of `Tue 26 May`


divdavem:
:ok: 11d3afc80ac82e9a210d822f3312bf3b0e04f3f8 seems fine for me.

PR #1472, opened 2015-05-27
fix Increase the default test timeout for Preselect autocomplete or multiautocomplete based test cases

fbasso:
As these tests are quite long, some browsers may failed (it happened on firefox) with timeout while the test is still rightly running.

Increasing the default time out for these tests is safe and avoid false failure.


divdavem:
:ok: 35e8e64c35e486a7ee5d4b0a3d5f52fe291c6794 seems fine for me.

fbasso:
Landed in 1f6be79

PR #1474, opened 2015-06-01
fix Better dialog width management depending on the title length

fbasso:


test/aria/widgets/container/dialog/sizes/DialogMaxWidthTestCase.js
```diff
@@ -0,0 +1,81 @@
...
+
+            var dialogIds = ["dialog1", "dialog2", "dialog3", "dialog4"];
+            this.waitFor({
+                condition : function() {
+
+                    // Wait for every dialog ready and open
+                    var dialogs = this.dialogs = {};
+                    for (var i = 0; i < dialogIds.length; i += 1) {
+                        var id = dialogIds[i];
+                        var currentInstance = dialogs[dialogIds[i]] = this.getWidgetInstance(dialogIds[i]);
```
divdavem: It would be better to use `id` instead of `dialogIds[i]`, otherwise, what's the point of creating the `id` variable?

test/aria/widgets/container/dialog/sizes/DialogMaxWidthTestCase.js
```diff
@@ -0,0 +1,81 @@
...
+                        if (!currentInstance) {
+                            return false;
+                        }
+                    }
+
+                    for(var id in dialogs) {
+                        if (!dialogs[id]._popup.isOpen) {
+                            return false;
```
divdavem: I think it would be better to include this check in the previous loop, instead of looping twice on the same things. What do you think?

src/aria/widgets/container/Dialog.js
```diff
@@ -580,9 +580,45 @@ module.exports = Aria.classDefinition({
             var cfg = this._cfg;
             var getDomElementChild = ariaUtilsDom.getDomElementChild;
             this._domElt = this._popup.domElement;
-            this._titleBarDomElt = getDomElementChild(this._domElt, 0, true);
-            this._titleDomElt = getDomElementChild(this._titleBarDomElt, cfg.icon ? 1 : 0);
+            var titleBarDomElt = this._titleBarDomElt = getDomElementChild(this._domElt, 0, true);
+            var bodyElement = getDomElementChild(this._domElt, 1, true);
```
divdavem: I would do: `var bodyElement = getDomElementChild(this._domElt, 0);` because the body element is not always the element before the last element inside `this._domElt`, especially if the dialog is resizable (cf https://github.com/fbasso/ariatemplates/blob/dialog_maxwidth/src/aria/widgets/container/Dialog.js#L315). However, it should always the first element element in `this._domElt`.

PR #1478, opened 2015-06-04
fix Forgotten use case after #1474

fbasso:


divdavem:
:ok: 0ccafbeb9cf72b039340f092cfe68e2844748738 seems fine for me

PR #1479, opened 2015-06-04
Fixed the setup of the deprecation for aria.utils.Device (and replicated for aria.core.Browser).

ymeine:
Deprecated methods which log a warning and then call the new implementation don't depend on the ability of the platform to provide property descriptors. I fixed the code according to that statement.

I also did an unrelated typo fix.


divdavem:
:ok: c88bbd19e073e3595317327cf30d6768e47d0982 seems fine for me if the travis build is ok.

PR #1480, opened 2015-06-05
fix test.aria.widgets.form.datepicker.errorstate.DatePicker fails on FF

fbasso:
The test is quite long. Increasing the default test timeout solve the issue


divdavem:
:ok: 03822b31089a05f6d5c1711935aab1902e98e0ed seems fine for me

PR #1483, opened 2015-06-05
fix regression after #1474, for test.aria.widgets.container.dialog.sizes.DialogWidthTestCase in IE7

fbasso:


divdavem:
:ok: 3026eee86333c686645a5b37ef41e8f5a6d1a0f5 seems fine for me

PR #1484, opened 2015-06-05
atbuild task modifies IO.swf

flongo:
When using `atbuild` grunt task, file IO.swf is modified, thus causing cross-domain requests to fail in packeged mode.

### DIFF
--- a/test/node/gruntTasks.js
+++ b/test/node/gruntTasks.js
@@ -22,6 +22,7 @@ var getPath = function (end) {
 };
 
 var testProjectPath = getPath("../nodeTestResources/testProject/");
+var srcPath = getPath("../../src/");
 
 var testBuild = function (index, callback) {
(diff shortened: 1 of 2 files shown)


test/node/gruntTasks.js
```diff
@@ -41,6 +42,20 @@ var testBuild = function (index, callback) {
 
 };
 
+var veryfyIOswfFile = function () {
+    var originalFile = fs.readFileSync(srcPath + "aria/resources/handlers/IO.swf", "utf8");
```
divdavem: A binary file should probably not be read in `"utf8"` mode. I think it is better to remove the second parameter to have a buffer as a return value, and use the `equals` method to compare them.

divdavem:
:ok: 6984eb63497d12c666248f7eecbbb83cd4bb54da seems fine for me.

PR #1488, opened 2015-06-10
Adding a cutYear option to the date parsing functions

divdavem:
This PR adds a `cutYear` option to the date parsing functions, which allows to configure the 2-digits number above which a 2-digits year 'xx' is considered '19xx' instead of '20xx'.


ymeine:
Good for me!

divdavem:
@ymeine Thank you for your review.

PR #1490, opened 2015-06-18
[WAI-ARIA] Adding the waiAria flag for widgets in the AriaLib

divdavem:
The `waiAria` flag allows to enable or disable accessibility either globally or at each widget level.


benouat:
@divdavem This change is fine for me.
We can merge it. :+1:

divdavem:

divdavem:
For future reference, see also #1494 and the associated commit 3cdfb9f9b8358a98b761371c132d8a6aa221f140.

PR #1491, opened 2015-06-19
[WAI-ARIA] aria:AutoComplete widget

divdavem:
This PR adds the DOM attributes required for the aria:AutoComplete widget to comply with the WAI-ARIA specifications (when the `waiAria` flag is enabled).

It is built on top of #1490

### DIFF
--- a/src/aria/widgets/form/list/CfgBeans.js
+++ b/src/aria/widgets/form/list/CfgBeans.js
@@ -99,6 +99,15 @@ module.exports = Aria.beanDefinitions({
                     $description : "strict: for strict highlighting (exact match only), always: for selecting the first item everytime, none: for no highlighting",
                     $enumValues : ["strict", "always", "none"],
                     $default : "none"
+                },
+                "waiAria" : {
+                    $type : "json:Boolean",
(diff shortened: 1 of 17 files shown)


src/aria/widgets/form/list/CfgBeans.js
```diff
@@ -99,6 +99,15 @@ module.exports = Aria.beanDefinitions({
                     $description : "strict: for strict highlighting (exact match only), always: for selecting the first item everytime, none: for no highlighting",
                     $enumValues : ["strict", "always", "none"],
                     $default : "none"
+                },
+                "waiAria" : {
+                    $type : "json:Boolean",
+                    $description : "If true, accessibility-related DOM attributes are enabled, to comply with WAI-ARIA specifications. This allows screen readers and other accessibility tools to work better."
+                },
+                "listItemDomIdPrefix" : {
```
benouat: @divdavem Do you really think this is needed to open such an api ? we could figure out something automagically....

benouat:
@divdavem apart from my comment that I don't have yet a better solution, it is fine for me right now.
I'll get back to you as soon as i have an idea !

divdavem:
@benouat Thank you

PR #1492, opened 2015-06-19
Fixing regression with templates depending on server resources

divdavem:
This commit fixes a regression introduced in Aria Templates v1.7.6 (dc9c3ba89fdaaf5a67fc2707f916d1f599b4a56d) :
templates which have a server resources dependency (i.e. whose classpath ends with `.Res`) failed to be loaded.

For example, if the template contained something like this:
```
 {Template {
   $classpath : "x.y.MyTemplate",
   $res : {
     myRes : "x.y.myResourceName.Res"
   }
 } ...
```
That template could not be loaded, with the following error message:
```
 [aria.core.MultiLoader] Load failed
 ReferenceError: myResourceName is not defined
```


benouat:
Change is fine for me :+1: assuming that the build is fine :)

divdavem:
@benouat Thank you for your review. I have fixed the test, rebased the branch and integrated it in master.

PR #1493, opened 2015-06-19
cutYear option not used in interpretFullDateRef (cf #1488)

divdavem:
This PR adds support for the `cutYear` option in the case of `interpretFullDateRef` (missing case in #1488).


benouat:
Fine for me, @divdavem you can merge :+1:

divdavem:
@benouat Thank you :smiley:

PR #1494, opened 2015-06-22
[WAI-ARIA] Moving waiAria flag to appSettings (instead of widgetSettings)

divdavem:
With this PR, the `waiAria` flag is now inside `appSettings` so that it can be used outside `AriaLib` widgets (e.g. for processing indicator...).


benouat:
@divdavem fine for me !

divdavem:
@benouat Thank you

PR #1496, opened 2015-06-23
aria.widgets.wai.autoComplete.* tests failing

divdavem:
This PR fixes the aria.widgets.wai.autoComplete.* failing tests.
There were random errors due to some checks which could happen before the list widget was fully loaded.
There were also some errors on IE 7 because of the wrong implementation of the partial polyfill for querySelectorAll (also because of the missing hasAttribute method on IE 7).


fbasso:
02d04d6 is fine for me

divdavem:
@fbasso Thank you

PR #1497, opened 2015-06-26
fix test.aria.templates.resources.TemplateWithResourcesTestCase

divdavem:
This commit fixes the TemplateWithResourcesTestCase test when it is run with devMode = true.


simonarbuckle:
:ok: for me

PR #1499, opened 2015-07-01
Browser detection improvements: adding isEdge, fixing issues

divdavem:
This commit adds the new isEdge flag to allow Microsoft Edge to be detected. For this purpose, the version of `ua-parser-js` was updated, which also fixes an invalid name for the Android Browser on some devices, and simplifies the code regarding PhantomJS detection.

Note that this pull request does not include any fix to improve the compatibility of Aria Templates with Microsoft Edge (which will be added in a later PR, if required)

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,6 @@
     "grunt-verifylowercase": "0.2.0",
     "jade": "1.1.5",
     "mocha": "1.17.0",
-    "ua-parser-js": "0.7.3"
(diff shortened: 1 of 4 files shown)


ymeine:
This is fine for me, but let's wait a bit (before start of the next release) to see how far the rest of the framework can be tested in current Edge, and if a release of UAParser.js comes soon.

PR #1500, opened 2015-07-06
fix aria.pageEngine.utils.BaseNavigationManager mustn't throw an error if the storage is not available

fbasso:


src/aria/pageEngine/utils/BaseNavigationManager.js
```diff
@@ -26,6 +26,8 @@ module.exports = Aria.classDefinition({
     $classpath : "aria.pageEngine.utils.BaseNavigationManager",
     $statics : {
 
+        STORAGE_NOT_UNAVAILABLE : "The storage is not available on this browser.",
```
divdavem: I would name this constant either `STORAGE_UNAVAILABLE` or `STORAGE_NOT_AVAILABLE` but `STORAGE_NOT_UNAVAILABLE` is really confusing!

src/aria/pageEngine/utils/BaseNavigationManager.js
```diff
@@ -55,9 +57,28 @@ module.exports = Aria.classDefinition({
          */
         this._navigate = cb || null;
 
+        var storageAvailable = true;
```
divdavem: I think doing it the opposite way would be better: initialize the variable with `false`, then set it to true after successfully assigning `this._storage`. This way, you can remove the `else` branch below.
And, maybe, you could even remove the `storageAvailable` variable completely and just check `this._storage` in the `if` statement.

divdavem:
Do you think you could add a test?

src/aria/pageEngine/utils/BaseNavigationManager.js
```diff
@@ -55,8 +57,25 @@ module.exports = Aria.classDefinition({
...
+             * Used to store state information for page refresh and external navigation
+             * @type aria.storage.LocalStorage
+             * @private
+             */
+            try {
+                this._storage = new ariaStorageLocalStorage();
+                storageAvailable = true;
+            } catch(e) {
+                this._storage = null;
+                storageAvailable = false;
```
divdavem: This line `storageAvailable = false` is not needed.

divdavem:
:ok: bc4ebd4fb755826f8d0e0eff5d6a722ef5db495f seems fine for me.

PR #1502, opened 2015-07-09
Radio button and check box icon not centered vertically when the label has multiple lines

divdavem:
This pull request fixes an issue with the check box and radio button when the label is large enough to wrap over multiple lines: the icon was not centered vertically.


fbasso:
72f8c94 is fine for me

divdavem:

PR #1503, opened 2015-07-10
fix aria.utils.Dom.scrollIntoView issue on Chrome

divdavem:
aria.utils.Dom.scrollIntoView failed to scroll the document on Chrome when the body had no margin.


fbasso:
f51a6a0 is fine for me

divdavem:

PR #1506, opened 2015-07-16
Adding WAI-ARIA to loading indicator

benouat:
This change introduces the generation of an `aria-live="polite"` attribute on the span that contains informative text on loading overlays.


divdavem:
:ok: ba45279a93b8e3c5e3c505d1875ae0523b3e9b39 seems fine for me.

PR #1511, opened 2015-07-27
Adding WAI support for the CheckBox widget

simonarbuckle:
The following attributes are now supported for the CheckBox:

* role
* aria-disabled
* aria-checked
* aria-label
* aria-labelledby

### DIFF
--- a/src/aria/widgets/Icon.js
+++ b/src/aria/widgets/Icon.js
@@ -17,7 +17,6 @@ var ariaWidgetsIconStyle = require("./IconStyle.tpl.css");
 var ariaWidgetsWidget = require("./Widget");
 var ariaCoreTplClassLoader = require("../core/TplClassLoader");
 
-
 /**
  * Aria Icon Widget
  */
@@ -55,6 +54,12 @@ module.exports = Aria.classDefinition({
          * @type String
          */
         this._cssClassNames = "xWidget";
+
+        /**
+         * Extra attributes for the widget.
+         * @type String
+         */
+        this.extraAttributes = "";
     },
     $statics : {
         // ERROR MESSAGES:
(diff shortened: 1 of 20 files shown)


src/aria/widgets/Icon.js
```diff
@@ -206,7 +212,8 @@ module.exports = Aria.classDefinition({
             if (cfg) {
                 var domEvtWrapper;
                 if (domEvent) {
-                    domEvtWrapper = new aria.templates.DomEventWrapper(domEvent);
+                    domEvtWrapper = new aria.templates.DomEventWrapper(domEvent
+                    );
                 }
```
divdavem: There is an issue with your formatter.
Please remove any change which are only about formatting so that we can quickly see real changes.

src/aria/widgets/form/CheckBox.js
```diff
@@ -229,10 +243,12 @@ module.exports = Aria.classDefinition({
         _onBoundPropertyChange : function (propertyName, newValue, oldValue) {
             if (propertyName === 'value') {
                 this._cfg.value = newValue;
+                this._updateWAIAttributes('aria-checked', newValue);
```
divdavem: This line can be removed, as it is also done in `_updateFomForState()` which is called line 248.

src/aria/widgets/form/CheckBox.js
```diff
@@ -229,10 +243,12 @@ module.exports = Aria.classDefinition({
...
                 this._setState();
                 this._updateDomForState();
             } else if (propertyName === 'disabled') {
                 this._cfg.disabled = newValue;
+                this._updateWAIAttributes('aria-disabled', newValue);
```
divdavem: For the sake of consistency, I would also move this line to `_updateDomForState`.

src/aria/widgets/Widget.js
```diff
@@ -367,6 +374,10 @@ module.exports = Aria.classDefinition({
                 });
             }
 
+            if (this._extraAttributes.length > 0) {
```
divdavem: You could remove the `if` statement.

divdavem:
:ok: a9565573fa75dbda036ac9720b37e7cbeea330a9 seems fine for me.

PR #1513, opened 2015-07-28
Allowing dialogs to be inside a container

divdavem:
This PR adds a new `container` property for the `@aria:Dialog` widget.
If this property is defined, the dialog is inserted as a child of the corresponding DOM element (which must have the `position:relative` or `position: absolute` style), and the dialog can only be moved or resized inside this element. If the dialog is modal, the modal mask is only above the container, and focusing elements outside the container is still possible.
This PR also contains a commit which fixes issues in the Resize/Drag utilities.


benouat:
@divdavem  this PR is fine for me. Don't squash the commits to push to master. Leave it like that.

divdavem:

PR #1514, opened 2015-07-30
fix The 204 code was not managed for a json response

fbasso:


divdavem:
:ok: 6e1766731ea1eb45c4773ff5dc3dafea8b726c26 seems fine for me, but you should add a test to check we do not introduce regressions later.

divdavem:
:ok: 83968597399ef5b90d67c2eb6265dbd15770dd15 seems fine for me.

PR #1516, opened 2015-08-04
feat Possibility to use Microsoft7MapProvider with https

flongo:


divdavem:
:ok: 06475bb0af1594cacb302c63ef73431f1ee51dff is fine for me. I have just integrated it.

PR #1520, opened 2015-08-13
Adding DialogMaxWidthTestCase to a test suite

divdavem:


divdavem:
The test fails on travis but not on my Windows machine! It seems that travis and my machine are using different fonts to display the dialog title which results in different dialog widths when the width can freely follow the width of the title.

fbasso:
I closed it, as the test as been added in the test suite in this commit : 4a75ce0

PR #1522, opened 2015-08-17
fix #1521 CheckBox not appearing when using the ATFlatSkin.

### LINKED ISSUE #1521: Checkbox not appearing when using the ATFlatSkin in IE11
In IE11, when using the ATFlatSkin the checkbox sprite is not loaded.  The sprite is being loaded via a data uri which embeds the image data for the background-image in the css for the skin.

simonarbuckle:
The skin was using a data uri to embed the image data within the background-image of the css for the skin.  This was causing problems with IE11 and as a result wasn't loading the sprite for the CheckBox.


divdavem:
:ok: 799973cd876d6fffd11d1d34f769c92b5866bb8d seems fine for me

divdavem:
Maybe you could update the commit message as files are now in the gif format.

PR #1523, opened 2015-08-19
fix aria:Link disabled color when skin contains multiple `sclass`

divdavem:
This commit fixes the following issue: when different `sclass` with different colors for the disabled state are defined for the aria:Link widget, all disabled links were using the same color for the disabled state, regardless of the `sclass` property.


fbasso:
bee9cc2 is fine for me

divdavem:

PR #1524, opened 2015-08-21
Empty label issue with aria:SelectBox widget

divdavem:
This pull request fixes the following issue: when a (non-null) value of an aria:SelectBox widget is selected and has an empty label, clicking on the field and immediately leaving it changes the value to null. On IE11 this also happens when clicking on the dropdown button.


fbasso:
fbb07b6 is fine for me

divdavem:

PR #1525, opened 2015-08-24
fix DialogMaxWidthTestCase refactor

fbasso:
The refactor has been done in order to face some skin and browser issues (font not the same, the test failed with phantom).

The commit also includes the test in the test suite, as it was missing (see PR #1520)


ymeine:
Good for me, you can merge as soon as the build passes.

PR #1526, opened 2015-09-03
Adding skin properties for the helpText

divdavem:
This PR adds some skin properties which allow to change the font and position of the help text in TextInput-based widgets.


fbasso:
b48be28 is fine for me

PR #1527, opened 2015-09-08
Issues with the sortKeyGetter property of the SortIndicator widget

divdavem:
This PR fixes the following issues with the sortKeyGetter property of the SortIndicator widget:
- it was not possible to use an object with `scope`, `fn` and `args` properties
- it could not be `null`, to disable the automatic sort done by the widget


fbasso:
043e157 is fine for me

divdavem:

PR #1529, opened 2015-09-09
fix Reduce the default waitFor timeout to increase speed tests

fbasso:


divdavem:
:ok: 95d75859ffd684d9ef77a8bc664cba1381b0020f seems fine for me

PR #1530, opened 2015-09-09
MultiSelect does not prevent default action with arrow down key

divdavem:
This PR makes sure the default action when pressing the arrow down key (usually: scrolling the document) is prevented when the focus is inside the aria:MultiSelect widget.


fbasso:
9202f9a is fine for me

divdavem:

PR #1531, opened 2015-09-14
Bridge url map and root map

### LINKED ISSUE #1074: aria.utils.Bridge does not forward the UrlMap and RootMap when opening a window
As a result, a packaged application can not load its own classes.

divdavem:
This PR fixes #1074.
Unlike #1075, it links the url map and root map of the popup window to the ones of the parent window, so that if the maps of the parent window are later updated, the popup window benefits from those changes. It also contains a test.


fbasso:
2e2fc9b is fine for me

divdavem:
@fbasso  Thank you for your review

PR #1532, opened 2015-09-14
Logging error for classpath dependencies outside the old loader

divdavem:
This PR adds an error in the console in case classpath dependencies are used outside the backward-compatible loader.

The backward-compatible loader is automatically used to load a class when no call to "require" is done in the module being loaded.

However, using both "require" and old-style classpath dependencies in a module can cause issues, because, in CommonJS, a module is considered loaded as soon as its content has been executed, but this is too early if classpath dependencies still have to be loaded. As a consequence, another class can be executed when its dependencies are not all fully loaded.

So this commit helps the developer to detect such issues and to fix them as early as possible.


benouat:
@divdavem I am :ok: with this change, as long as the build finishes, you can merge.

divdavem:

PR #1537, opened 2015-09-21
fix FocusDialogContainerTestCase on Edge (after reducing waitFor delay)

divdavem:
This commit adds some pauses in the test because of the following strange behavior in Edge: clicking on another field just after focusing a field with the keyboard has no effect to change the focus, it is necessary to wait a bit after focusing the field before the click on another field has some effect.

### DIFF
--- a/test/aria/widgets/container/dialog/container/FocusDialogContainerTestCase.js
+++ b/test/aria/widgets/container/dialog/container/FocusDialogContainerTestCase.js
@@ -55,7 +55,8 @@ Aria.classDefinition({
             }
 
             function step8() {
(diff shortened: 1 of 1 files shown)


divdavem:
The issue seems to be due to a touchpad setting which can be disabled in Windows 10:
Settings -> Devices -> Mouse & touchpad -> Touchpad section

> To help prevent the cursor from accidentally moving while you type, change the delay before clicks work

The setting can be set to: `No delay (always on)`

divdavem:
I am closing this issue as it can be better solved by changing the setting in Windows 10.

PR #1533, opened 2015-09-15
Adding WAI support for the Radio widget

fbasso:

PR #1538, opened 2015-09-22
fix The checkbox can't be checked after commit #1533

fbasso:


divdavem:
:ok: a0c8a9ec8d6bc15e45611307ed857735376e909e seems fine for me

PR #1539, opened 2015-09-23
waitForDropDownPopup in TemplateTestCase and making AdaptToContentWidthTest more reliable

divdavem:
This PR adds the `waitForDropDownPopup` method in `TemplateTestCase` and uses it in `AdaptToContentWidthTest` so that this test is more reliable.


fbasso:
b484d17 is fine for me

divdavem:

PR #1543, opened 2015-10-08
fix #1542 issue with AutoComplete display text on iPad

### LINKED ISSUE #1542: AutoComplete display value incorrect on iPad
Tested with an iPad 2 with iOS 8.3:

Occurs when an AutoComplete already has display text and is configured to select that text on tap.  

* The issue:
If a user taps the AutoComplete field, the display text is selected, then when the user taps a character from the virtual keyboard the new character is appended to the existing display text instead of replacing it.

simonarbuckle:


fbasso:
39fde06 is fine for me, as far as Travis succeed.

PR #1544, opened 2015-10-09
Fix random issue in preselectAutofill tests

divdavem:
This PR fixes random issues in preselectAutofill tests due to the fact that the condition to wait for the popup was not complete enough. Using the dedicated `waitForDropDownPopup` method (from `TemplateTestCase`, added in 58c8e7a0402b8e97251e46793b440e4c6021defd) solves the issue.


benouat:
@divdavem :+1: for me.

divdavem:

PR #1545, opened 2015-10-23
Updating attester (2.4.0)

divdavem:


benouat:
@divdavem can't we keep multiple versions of node in the conf ? Maybe it might not be that useful... I am just wondering

divdavem:
We could indeed, but it would then run the test suite once for each node.js version. I don't think it is really useful, unless you have some specific idea in mind.

PR #1547, opened 2015-10-26
fix test.aria.utils.Dom on Opera

divdavem:


fbasso:
3a2abda is fine for me

divdavem:

PR #1549, opened 2015-10-28
fix #1546 adds the ability to remove appenders

### LINKED ISSUE #1546: Add removeAppender method in aria.core.Log
It should be possible to remove individual appenders, currently you can only clear all appenders.

simonarbuckle:


src/aria/core/Log.js
```diff
@@ -255,14 +302,26 @@ module.exports = Aria.classDefinition({
...
+         * same methods as aria.core.log.DefaultAppender)
          */
-        clearAppenders : function () {
-            var apps = this.getAppenders();
+        clearAppenders : function (appender) {
+            var apps = appender ? [appender] : this.getAppenders().slice(0);
             for (var i = 0, l = apps.length; i < l; i++) {
                 apps[i].$dispose();
+                this._remove(this._appenders, apps[i]);
             }
```
divdavem: By implementing the right loop in `clearAppenders`, we can avoid duplicating the code from `aria.utils.Array`. Just loop over all items (instead of only the item(s) to remove), and if the item is to be removed (either because there is no argument or because the item is the provided argument), remove it from the array with splice and dispose it. We don't need `_indexOf`, `_remove` and `_removeAt`.

divdavem:
:ok: b65d3cc9b721ebea07f94a89fc32d05e85c3cbfc seems fine for me

PR #1550, opened 2015-10-28
Adding isOpera whenever there is isChrome

divdavem:
The behavior in Opera is very close to the one of Chrome as both browsers have much code in common.


divdavem:
This PR fixes test.aria.widgets.container.dialog.indicators.DialogTestCase on Opera.

fbasso:
d4cbffc is fine for me, as local testing on Opera was fine.

divdavem:

PR #1551, opened 2015-10-28
Making WaiAutoCompleteTestSuite more robust

divdavem:
This PR makes the `WaiAutoCompleteTestSuite` test suite more robust by using the `waitForDropDownPopup` method instead of using an incomplete condition in a call to the `waitFor` method.


fbasso:
238e5c8 is fine for me

divdavem:

PR #1552, opened 2015-11-12
Fix wrong display of ampersand in the suggestions of the autocomplete

divdavem:
This PR fixes the following issue: when typing an ampersand in the autocomplete, and if some suggestions match, suggestions are wrongly displayed with '&' replaced by the corresponding '&amp;amp;' entity. Adding another letter afterwards switches back to the correct display, but then the highlighting in the suggestion no longer works (while there is an ampersand in the input).


benouat:
@divdavem this PR is fine for me. :+1:

divdavem:

PR #1553, opened 2015-11-19
fix: scrollBarX and scrollBarY were not taken into account in tooltip

divdavem:


benouat:
@divdavem just wait for CI, but the change is :ok: to me.

divdavem:

PR #1554, opened 2015-11-27
feat Allow aria-* attributes

flongo:


benouat:
@flongo I am fine with this change. @divdavem what do you think ?

test/aria/utils/Html.js
```diff
@@ -51,6 +51,14 @@ Aria.classDefinition({
                                 this.assertEquals(got, value, "data-" + dataKey + " should be %2, got %1");
                             }
                         }
+                    } else if (key === "aria") {
+                        for (var ariaKey in attribute) {
+                            if (attribute.hasOwnProperty(ariaKey)) {
+                                var value = stringUtil.encodeForQuotedHTMLAttribute(attribute[ariaKey]);
```
divdavem: There should be no call to `encodeForQuotedHTMLAttribute` here (it was also wrong line 46), but, anyway, the test does not check attributes containing quotes, so the encoded version is identical to the non-encoded version.

flongo: It's that in general this test re-implements the code itself, I am not a big fan...

divdavem:
@flongo Thank you for taking my comment into account.
537604610a69cb559821fc60d6007196b2f75df7 seems fine for me

PR #1556, opened 2015-12-04
Adding WAI-ARIA accessibility support to the DatePicker widget

divdavem:
This PR adds accessibility attributes on the DatePicker DOM elements (`role`, `aria-activedescendant`, `aria-selected`, `aria-haspopup`, `aria-expanded`, `aria-label`...) when the `waiAria` widget property is `true`.
Also, no matter what the value of `waiAria` is, the browser focus is now always on the Calendar DOM element when the popup is open, instead of being on the dropdown button.


fbasso:
245a9f4 is fine for me

divdavem:

PR #1557, opened 2015-12-04
Fix issue when trying to close the AutoComplete popup by clicking on an iframe

simonarbuckle:


divdavem:
:ok: f123c7a173b77fcd627f68089da702c8b57fc2bf seems fine for me.

PR #1567, opened 2015-12-21
fix #1566 alphabetic characters are not accepted as valid entries

### LINKED ISSUE #1566: TimeField widget
In the TimeField widget, entering alphabetic characters is sometimes accepted rather than flagged as an error:

- 12b is accepted and shown as 12:00
- b12 is flagged as an error
- 1xx1 is accepted and shown as 11:00
- xx is flagged as an error

simonarbuckle:
Previously there were cases where some alphabetic characters were accepted as valid entries and subsequently transformed into valid times, now all alphabetic characters are illegal unless they are one of the following special cases:

- a m
- p m
- h
- m

### DIFF
--- a/src/aria/utils/Date.js
+++ b/src/aria/utils/Date.js
@@ -122,7 +122,7 @@ var parseNextSteps = function(state, options) {
     datetime = currentState.datetime;
 
     var matches = nextPatternRegexp.exec(state.pattern);
-    //debugger;
+
     if (matches) {
         var fullPattern = matches[1];
         var patternType = fullPattern.substr(0, 1);
@@ -492,6 +492,7 @@ var ariaUtilsDate = module.exports = Aria.classDefinition({
             scope : this
         });
 
+        this._validateEntry = /^(?:(?![a-z])\d\s?[\\h;m,.\-:\/\s]?\s?\d?\s?(am|pm|a\.m|p\.m|a\sm|p\sm|a.m.|p.m.|a. m|p. m|a .m|p .m|a . m|p . m)?\s?)+$/;
     },
     $destructor : function () {
         this._formatCache = null;
@@ -558,8 +559,8 @@ var ariaUtilsDate = module.exports = Aria.classDefinition({
 
             entry = entryStr;
 
-            // need to check for at least one digit at the beginning of the string
-            if (!this._isValidTime(entry)) {
+            // need to check for alphabetic characters
+            if (!this._validateEntry.test(entry)) {
                 return null;
             }
 
@@ -775,12 +776,8 @@ var ariaUtilsDate = module.exports = Aria.classDefinition({
             var characterCheck1 = time.charAt(0);
             var characterCheck2 = time.charAt(1);
 
-            // where the first two digits are 24 need to set the hour to 00
             if (characterCheck1 === '2' && characterCheck2 === '4') {
-                hours[0] = 0; // value to be used for hours
-                hours[1] = 2; // number of digits for hours in entry
-                // string
-                return hours;
+                return null;
             }
 
             // where there is a second digit and (the first digit is 2 the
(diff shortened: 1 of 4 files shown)


src/aria/utils/Date.js
```diff
@@ -492,6 +492,7 @@ var ariaUtilsDate = module.exports = Aria.classDefinition({
             scope : this
         });
 
+        this._validateEntry = /^((?![a-z])\d(\s)?([\\h;m,.\-:\/\s])?(\s)?(\d)?(\s)?(am|pm|a\.m|p\.m|a\sm|p\sm|a.m.|p.m.|a. m|p. m|a .m|p .m|a . m|p . m)?(\s)?)+$/;
```
ymeine: Please turn capturing groups to non-capturing groups by adding `?:` right after the first parenthesis.
Also, removing unnecessary groups: `\s`, `\d` `[...]` are all atomic and the quantifiers behind them don't require a group in this case.
This will ease the reading of the expression.

test/aria/utils/Time.js
```diff
@@ -142,10 +132,10 @@ Aria.classDefinition({
          */
         test_interpretTime : function () {
             var ariaTimeUtil = aria.utils.Date;
-            var useCases = ["22:4p:1", "99", "9p", "9h00", "9:9", "11:09   p", "9 9", "9/2", "9;2", "9-2", "9:9",
+            var useCases = ["99", "9h00", "9:9", "9 9", "9/2", "9;2", "9-2", "9:9",
```
ymeine: I now that part was already present, and also that a future cleaning is planned. But this isn't a lot to do: can you change the entries to object, with two properties, one holding the input value (former `useCases` entry) and one holding the expected result (former `checker` entry).
This would be fare more logical.

Example: 

```javascript
var useCases = [{input: "99", time: "09:09:00"}];
```

test/aria/widgets/form/timefield/checkValue/TimeField.js
```diff
@@ -19,16 +19,115 @@ Aria.classDefinition({
...
+            this.check4Digits();
+            this.checkAlphabetic();
+            this.checkValuesWithAMPM();
+        },
+        check1Digit : function () {
+            var hour, result, i;
+
+            // 1 digit = [h]
+            for (i = 0; i < 10; i++) {
+                hour = i + '';
```
ymeine: I would reverse the order of operands here: it would make it more obvious we are casting to a string (and it seems to be more widespread).

test/aria/widgets/form/timefield/checkValue/TimeField.js
```diff
@@ -19,16 +19,115 @@ Aria.classDefinition({
...
+            this.checkValues.push(result);
+            this.checkValuesWithSeparator(value);
+        },
+        checkAlphabetic : function () {
+            // alphabetic characters are invalid
+            this.testValues.push("12b", "b12", "1xx1", "xx");
+            this.checkValues.push("12b", "b12", "1xx1", "xx");
+
+        },
+        checkValuesWithAMPM : function () {
```
ymeine: I would create two distinct loops (despite the fact that in practice the two arrays have the same length).

Also I would rename some variables: 

- `items` :arrow_right: `index`
- `valid` :arrow_right: `baseInput` or something similar (you are building an also valid value with it)

test/aria/widgets/form/timefield/checkValue/TimeField.js
```diff
@@ -19,16 +19,115 @@ Aria.classDefinition({
...
+                // valid AM
+                this.testValues.push(valid + this.am[items]);
+                this.checkValues.push("01:00");
+
+                // valid PM
+                this.testValues.push(valid + this.pm[items]);
+                this.checkValues.push("13:00");
+            }
+        },
+        checkValuesWithSeparator : function (value) {
```
ymeine: You could lighten the code by using the square brackets notation instead of the `charAt` method. Of course in this case don't forget to convert the value to an array before to remain compatible with IE7: `value.split('')`.

ymeine:
The new behavior is fine for me. The code style will be improved through the upcoming cleaning task.

simonarbuckle:
@ymeine thanks for your review

PR #1568, opened 2016-01-05
fix Checkboxes and Radios didn't work on IE with JAWS activated

fbasso:
Space/Enter had no effect.

AT-1242


src/aria/widgets/form/CheckBox.js
```diff
@@ -124,21 +116,19 @@ module.exports = Aria.classDefinition({
...
-                        cfg._inputType, '"', name, ' value="', cfg.value, '"/>'].join(''));
             }
+            var tabIndex = this._cfg.tabIndex;
+            tabIndex = !cfg.disabled ? 'tabindex="' + this._calculateTabIndex() + '" ' : "";
+
+            out.write(['<input', cfg.disabled ? ' disabled' : '',
+                    this._skinObj.simpleHTML ? ' style="display:inline-block"' : ' class="xSROnly"',
+                    (Aria.testMode || cfg.waiAria) ? ' id="' + this._domId + '_input"' : '',
+                    this._isChecked() ? ' checked' : '', ' type="', cfg._inputType, '"', name, ' value="',
+                    cfg.value, '" ', tabIndex, this._waiAriaAttributes, '/>'].join(''));
```
divdavem: You should probably remove `this._waiAriaAttributes` as it is not used anywhere.

src/aria/widgets/form/CheckBox.js
```diff
@@ -104,8 +96,8 @@ module.exports = Aria.classDefinition({
          */
         _checkCfgConsistency : function () {
             // for the simpleHTML case, tab index is set directly on the input
```
divdavem: This comment should be updated. Now the tab index is always set directly on the input.

src/aria/widgets/form/CheckBox.js
```diff
@@ -104,8 +96,8 @@ module.exports = Aria.classDefinition({
...
-            this._customTabIndexProvided = this._skinObj.simpleHTML;
-        },
+            this._customTabIndexProvided = true;
```
divdavem: I think we don't really need to define the `_checkCfgConsistency` method. We could move this line to the constructor, or even set the `_customTabIndexProvided` property directly in the prototype.

src/aria/widgets/form/CheckBox.js
```diff
@@ -66,14 +66,6 @@ module.exports = Aria.classDefinition({
          * @type Boolean
          */
         this._mousePressed = false;
```
divdavem: You could completely remove `this._mousePressed` as it is not used anymore anywhere.

test/aria/widgets/wai/input/WaiInputTestSuite.js
```diff
@@ -28,7 +28,7 @@ Aria.classDefinition({
         this.addTests("test.aria.widgets.wai.input.multiSelect.MultiSelectTestCase");
         this.addTests("test.aria.widgets.wai.input.datePicker.DatePickerTestCase");
         this.addTests("test.aria.widgets.wai.input.dateField.DateFieldTestCase");
-        this.addTests("test.aria.widgets.wai.input.checkBox.WaiCheckBoxTestSuite");
-        this.addTests("test.aria.widgets.wai.input.radiobutton.WaiRadioButtonTestSuite");
```
divdavem: Tests which are no longer passing and which are not relevant should be removed (including test files), and not only commented in the test suite.

divdavem:
:ok: 39e5ad44dcfa8383b8e976b8763245f071451484 seems fine for me

PR #1569, opened 2016-01-06
span elements made focusable by Firefox when `overflow-y:hidden` is used alone

divdavem:
This PR makes sure we always specify both `overflow-x` and `overflow-y` (or none of them) so that Firefox does not make the corresponding &lt;span&gt; element focusable when it is not supposed to be focusable.


fbasso:
22c64bb is fine for me

divdavem:

PR #1555, opened 2015-12-01
Dynamic zIndex popup management

divdavem:
When clicking on a popup, or focusing a widget inside it, the popup is now brought to the front.

PR #1570, opened 2016-01-08
Fix regressions caused by the dynamic zIndex popup management feature

divdavem:
This commit fixes regressions introduced by commit 8fed6379173c4bfbc48b2d2d962bd197d8edde9a (#1555): if a dialog contains widgets which have a popup (such as the dropdown list of an `@aria:Select`, or the validation error popup of most widgets) the popup is often displayed behind the dialog.
This is fixed by adding a new `zIndexKeepOpenOrder` property on popups, which, when set to `true` (which is the default), makes sure the newly opened popup is always kept in front of any popup that was opened before.
This property is set to `false` on all dialogs, to make sure they can be reordered as needed.

### DIFF
--- a/src/aria/popups/Beans.js
+++ b/src/aria/popups/Beans.js
@@ -135,6 +135,11 @@ module.exports = Aria.beanDefinitions({
                     $description : "When the popup is being opened, the animation is applied",
                     $sample : "slide left"
                 },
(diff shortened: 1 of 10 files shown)


benouat:
@divdavem changes are :ok: for me.

divdavem:

PR #1571, opened 2016-01-12
Making the MSDownArrowKey test use waitForDropDownPopup

divdavem:
This PR makes the `test.aria.widgets.form.multiselect.downArrowKeyPreventDef.MSDownArrowKey`
test more reliable by using the common `waitForDropDownPopup` method instead of re-implementing it in an incomplete way.


fbasso:
d28d31e is fine for me

divdavem:

PR #1572, opened 2016-01-13
fix #1572 Texttfield autoselect doesn't work with JAWS and the arrow keys

fbasso:
- Autoselect won't work when the textfield was reached with the keyboard

### DIFF
--- a/src/aria/widgets/form/TextInput.js
+++ b/src/aria/widgets/form/TextInput.js
@@ -911,7 +911,6 @@ module.exports = Aria.classDefinition({
          * @protected
          */
         _dom_onclick : function () {
-            this._autoselect();
             if (!!this._cfg.onclick) {
                 this.evalCallback(this._cfg.onclick);
             }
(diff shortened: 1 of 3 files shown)


divdavem:
Now that you call `this._autoselect()` from `_dom_onfocus`, I think it would be possible to remove `this._autoselect()` from [line 1271](https://github.com/fbasso/ariatemplates/blob/wai_textfield/src/aria/widgets/form/TextInput.js#L1271) in the `focus` method:

```js
            if (!fromSelf) {
                // IE FIX: the focus() can be asynchronous, so let's add a timeout to manage the autoselect
                ariaCoreTimer.addCallback({
                    fn : function () {
                        this._autoselect();
                    },
                    scope : this,
                    delay : 25
                });
            }
```

divdavem:
:ok: d5cd72d6e7b78576f069e934d26f2e547f666de9 seems fine for me

However, the commit message should be changed before integrating this PR in the master branch:
- [ ] replace `autofocus` by `autoselect` in the commit message
- [ ] add `CR 08528144` on a new line at the end of the commit message.

PR #1573, opened 2016-01-14
WAI "required" must be used instead of "aria-required"

fbasso:


divdavem:
:ok: 5a3a163b13e3f6e5430db6abaee200a08fa75495 seems fine for me.
It would be good to add `CR 08528144` on a new line at the end of the commit message before integrating the commit in master.

PR #1574, opened 2016-01-15
Allowing to easily use attester-results-ui with the `npm run attest` command

divdavem:
This is a small change in the `npm run attest` script which is necessary to (easily) allow connecting attester-results-ui to the attester instance started by this script.


fbasso:
29c7c68 is fine for me

divdavem:

PR #1575, opened 2016-01-15
Allowing to write Jaws automatic tests

divdavem:
This commit adds the `aria.jsunit.JawsTestCase` base class to help writing tests to be run with the Jaws screen reader enabled.
It also adds the `test.JawsTestSuite` class which is intended to reference all test cases designed to be run with Jaws enabled.
It also includes a first Jaws test, which can also be used as an example to write further tests: `test.aria.widgets.wai.datePicker.DatePickerJawsTest1`


benouat:
@divdavem I am fine with that ! Great idea :smiley: :+1:

divdavem:

PR #1577, opened 2016-01-20
Introduction of a new cfg for the Dialog called focusIfModal which will

tomwaterer:
set the focus on the first focusable element when the Dialog is opened and
if and only if the dialog has the modal flag set to true

### DIFF
--- a/src/aria/widgets/container/Dialog.js
+++ b/src/aria/widgets/container/Dialog.js
@@ -609,9 +609,14 @@ module.exports = Aria.classDefinition({
 
             this._calculatePosition();
 
-            if (cfg.focusIfModal && cfg.modal) {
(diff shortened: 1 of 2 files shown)


src/aria/widgets/container/Dialog.js
```diff
@@ -609,7 +609,7 @@ module.exports = Aria.classDefinition({
 
             this._calculatePosition();
 
-            if (cfg.modal) {
+            if (cfg.focusIfModal && cfg.modal) {
```
fbasso: Hi,

If case of modal, and if the parameter is false, it would be nice to blur the active field by `Aria.$window.document.activeElement.blur()`. Otherwise, an active element under the mask can potentially stay focused under a modal.

PR #1578, opened 2016-01-20
fix #1578 validationDelay didn't work with dropdown based widgets

fbasso:


divdavem:
:ok: b4c083065ca8a5d021bfa4196bb8ac7b60e0d37e seems fine for me.

PR #1579, opened 2016-01-21
Now raising an event when bringing popups to the front

divdavem:
This is needed in case several popup managers need to cooperate, so that only one manages z-indexes.


fbasso:
8b8b4cf is fine for me

divdavem:

PR #1580, opened 2016-01-22
Now raising an event when preventing the focus because of a modal dialog

divdavem:
This is needed in case several popup managers need to cooperate, so that another popup manager can display a modal dialog above an Aria Templates modal dialog.


benouat:
I am fine with these changes.
@divdavem As soon as build is :ok_hand: you can merge it.

divdavem:
@benouat Thank you for your review.

PR #1581, opened 2016-01-26
fix Prevent SynEvent from crashing FF test on getSelection

fbasso:
On Firefox, selectionStart throws an error on an checkbox for instance (NS_ERROR_FAILURE)


divdavem:
:ok: 4901c06bdf7fabcc282d9d297c6f4b2efe2834f2 seems fine for me

PR #1582, opened 2016-01-26
Fixing test.aria.widgets.autoselect.programmatic.AutoSelect on IE9 and IE10

divdavem:
This PR fixes a regression introduced by commit f9d30b69f19e3d472f6b4c93ddbbb210776507d8.
On IE9 and IE10, setting the selection on the focus event handler itself does not work, it has to be delayed.


fbasso:
cbbfb2 is fine for me, as the tests are ok on your fork

divdavem:

PR #1583, opened 2016-02-03
No longer scrolls to the top of the dialog when closing a popup

divdavem:
This commit removes some code which was trying to focus the first element of a popup when another popup was closed. This behavior was disturbing as it could scroll to the top of the dialog in case the first element was not visible.


benouat:
@divdavem as shown to me on your laptop, the behavior is acceptable. You can deliver this change.

divdavem:

PR #1584, opened 2016-02-04
Using onreadystatechange to improve performance

divdavem:
This commit uses the onreadystatechange callback instead of the previous polling code to improve performance.


benouat:
@divdavem I am fine with these changes.

divdavem:

PR #1585, opened 2016-02-05
Issue when a resizable/movable dialog is disposed shortly after its creation

divdavem:
As the aria.utils.dragdrop.Drag and aria.utils.resize.Resize utilities are loaded asynchronously (only if the dialog configuration requires those classes), it is possible that the dialog is disposed before those
classes are loaded, so the callback should check it before doing anything else.


benouat:
@divdavem these changes are fine to me.

divdavem:

PR #1586, opened 2016-02-08
WAI-ARIA Improvements  for the AutoComplete

divdavem:
The compatibility of the aria:AutoComplete widget with JAWS on IE 11 is improved: it is now possible to specify a message to be read when the number of displayed suggestions changes and to change the message which is read for each suggestion.


benouat:
@divdavem should be fine for me. Let's deliver and wait for feedback !

divdavem:

benouat:
@divdavem don't forget to also perform changes in ariatemplates/documentation-code. We need a new sample.
Also remove the old one.

PR #1587, opened 2016-02-10
fix - increased (doubled) the timeout for the test of the task easypackage

ymeine:
I faced some issues with Travis builds where the timeout was just too low.


divdavem:
:ok: 944774fdc93e7bc7935dd7aa02dd896b2084fa2e is fine for me

PR #1592, opened 2016-02-12
feat WAI-ARIA - Modal dialog

ymeine:
Mainly brings accessibility to modal dialogs.

# Dialog features

## Label (title)

Added the label information for the popup associated to the modal dialog. It is bound to the element containing the text of the latter's title.

## Accessible icons

Features:

- labels for icons can be specified: they will result in tooltips and aria labels
	- this implied opening a new property for icons allowing the specification of a label (the tooltip one was already existing)
- icons are accessible using tab
- icons react on "return" and "enter" keys, to trigger their action
- the outline is displayed around icons when they are focused

Issues:

- the "initial" outline width for Firefox is too big and does not correspond to the actual native outline we can see on focused elements
- JAWS can't read properly the label of the first inserted icon (first in the DOM)

Glitches:

- the tab order for icons is visually reversed, due to the floating positioning being contrary to the DOM insertion order

## Background elements hiding

Features:

- the elements of the body that visually resides behind the opened modal dialog are now ensured as hidden as long as the dialog is opened. These elements are restored to their previous state when the dialog is closed.

Glitches:

- this feature implementation is limited to a sequence of modal dialogs that are closed in the reverse order of their opening.

## Focus restoration

Features:

- the element that was focused just before opening the dialog is memorized and the focus is put back on it when closing the dialog

Issues:

- the icon reacts on the key down event: if the element on which the focus is put back opens the dialog upon the counterpart key up event (same key), it makes it impossible to close the dialog without having it opened again immediately afterwards

Glitches:

- a simple reference to the native DOM element is internally saved: if this element instance is not present in the DOM anymore, this won't work (usually the case upon AT refreshes)

# Side changes

## fix - Action widgets (Button & Link)

Ensure keyup is handled only if keydown happened.

## feat - WAI-ARIA - Icon

Features:

- icons accept a new property to specify their labels
- icons are accessible using tab
- icons react on "enter" and "space" keys, to trigger their action
- the outline is displayed around icons when they are focused

Issues:

- the "initial" outline width for Firefox is too big and does not correspond to the actual native outline we can see on focused elements

## feat - MultiSelect

The escape key now closes the dropdown.

## fix - Fixed the escape behavior for several widgets

That way widgets such as the dialog can be closed with this key.

Now their associated controllers properly stop the behavior of the events as needed.

I had to add an additional property to the controllers' report model so that they can tell to prevent the default behavior of the browsers without making the propagation stop. This is needed to keep a bug fix while enabling the new behavior on escape.

### DIFF
--- a/src/aria/popups/Popup.js
+++ b/src/aria/popups/Popup.js
@@ -16,6 +16,7 @@ var Aria = require("../Aria");
 var ariaPopupsPopupManager = require("./PopupManager");
 require("./Beans");
 var ariaUtilsDom = require("../utils/Dom");
+var ariaUtilsString = require("../utils/String");
 var ariaUtilsSize = require("../utils/Size");
 var ariaUtilsEvent = require("../utils/Event");
 var ariaUtilsDelegate = require("../utils/Delegate");
@@ -320,10 +321,21 @@ module.exports = Aria.classDefinition({
                  ariaUtilsDelegate.getMarkup(this._delegateId),
                  ' style="position:absolute;top:-15000px;left:-15000px;visibility:hidden;display:block;"'
             ];
-            var role = cfg.role;
-            if (cfg.waiAria && role) {
-                html.push(' role="' + role + '"');
+
+            if (cfg.waiAria) {
+                var role = cfg.role;
+                if (role) {
+                    role = ariaUtilsString.escapeForHTML(role, {attr: true});
+                    html.push(' role="' + role + '"');
+                }
+
+                var labelId = cfg.labelId;
+                if (labelId) {
+                    labelId = ariaUtilsString.escapeForHTML(labelId, {attr: true});
+                    html.push(' aria-labelledby="' + labelId + '"');
+                }
             }
+
             html.push("></div>");
             div.innerHTML = html.join('');
 
@@ -690,13 +702,21 @@ module.exports = Aria.classDefinition({
          * @protected
          */
         _show : function () {
+            // --------------------------------------------------- destructuring
 
-            // Insure that the top left corner is visible
-            if (this.modalMaskDomElement) {
+            var conf = this.conf;
 
+            var domElement = this.domElement;
+            var modalMaskDomElement = this.modalMaskDomElement;
+
+            // ------------------------------------------------------ processing
+
+            // Ensure that the top left corner is visible
+            if (modalMaskDomElement) {
                 if (this._containerOverflow == -1) {
                     this._containerOverflow = this.popupContainer.changeContainerOverflow("hidden");
                 }
+
                 var containerSize = this.popupContainer.getScrollSize();
 
                 // Compute the style after scrollbars are removed from the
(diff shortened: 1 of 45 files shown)


src/aria/popups/Popup.js
```diff
@@ -320,10 +320,19 @@ module.exports = Aria.classDefinition({
...
+            if (cfg.waiAria) {
+                var role = cfg.role;
+                if (role) {
+                    html.push(' role="' + role + '"');
+                }
+
+                var labelId = cfg.labelId;
+                if (labelId) {
+                    html.push(' aria-labelledby="' + labelId + '"');
+                }
```
divdavem: Even if `role` and `labelId` are not supposed to contain quotes, I think it is good, to be sure we don't have injections, to either escape (with the utilities from `aria.utils.String`) `role` and `labelId` or to use `setAttribute` once the DOM element is available.

src/aria/popups/Popup.js
```diff
@@ -712,12 +729,23 @@ module.exports = Aria.classDefinition({
...
+                    'left:0px;',
+                    'top:0px;',
+                    'width:', containerSize.width, 'px;',
+                    'height:', containerSize.height, 'px;',
+                    'z-index:', modalMaskZIndex, ';',
+                    'position:absolute;',
+                    'display:block;'
+                ].join('');
+
+                if (conf.waiAria && conf.modal) {
```
divdavem: There is no need to check `conf.modal` here, because we are already in the `if (modalMaskDomElement) { ... }`.

src/aria/widgets/Icon.js
```diff
@@ -75,35 +74,92 @@ module.exports = Aria.classDefinition({
...
-            if (tooltip != null && tooltip !== '') {
-                tooltip = 'title="' + tooltip + '" ';
-            } else {
-                tooltip = '';
+            // ------------------------------------------------------ processing
+
+            var attributes = [];
+
+            function addAttribute(key, value) {
+                attributes.push(key + '="' + value + '"');
```
divdavem: The value should be escaped here too.

src/aria/widgets/action/Button.js
```diff
@@ -367,14 +364,16 @@ module.exports = Aria.classDefinition({
...
+                    this._updateState();
 
-                if (!this._performAction(domEvt)) {
-                    domEvt.stopPropagation();
-                    return false;
+                    if (!this._performAction(domEvt)) {
+                        domEvt.stopPropagation();
+                        return false;
+                    }
+                    return true;
```
divdavem: This `return true` is useless.

src/aria/widgets/action/Link.js
```diff
@@ -118,16 +117,22 @@ module.exports = Aria.classDefinition({
...
-                if (!this._performAction(domEvt)) {
-                    domEvt.stopPropagation();
-                    return false;
+                if (this._keyPressed) {
+                    this._keyPressed = false;
+                    if (!this._performAction(domEvt)) {
+                        domEvt.stopPropagation();
+                        return false;
+                    }
+                    return true;
```
divdavem: This `return true` is useless (as there is another one a bit further).

src/aria/widgets/controllers/reports/ControllerReport.js
```diff
@@ -36,6 +36,12 @@ module.exports = Aria.classDefinition({
         this.cancelKeyStroke = false;
 
         /**
+         * Controller specifies if the keystroke's default behavior has to be canceled
+         * @type Boolean
+         */
+        this.cancelKeyStrokeDefaultBahevior = false;
```
divdavem: It would be better to use the correct spelling for `Behavior` in the name of the property (this should be changed in all places where the variable is used).

src/aria/widgets/container/Dialog.js
```diff
@@ -682,8 +848,23 @@ module.exports = Aria.classDefinition({
          */
         close : function () {
             var cfg = this._cfg;
-            if (this._popup) {
+            var modal = cfg.modal;
+
+            if (modal && cfg.waiAria) {
```
divdavem: To better respect the DRY principle, instead of checking again the same condition as in the `open` method (with the risk that the condition differs after changing the code only in one place), I think it is better to only check here whether `this._hiddenElements` is set (and make sure it is set only in the `open` method with the right condition).

divdavem:
:ok: e597d567f5f645a1ccdd3bee358130e26b0a4200 seems fine for me. You can integrate it once the travis build is :ok:

PR #1593, opened 2016-02-12
Fix a random error in robot tests which use the backspace key in PhantomJS

divdavem:
The backspace key was not correctly sent to PhantomJS by `RobotPhantomJS` and this apparently triggered a navigation in the iframe to some previous test, resulting in a timeout.

Especially, for the Aria Templates test campaign, this commit fixes the random issue with `AdaptToContentWidthTest`, when it is run in the same
PhantomJS instance as `ExternalHashNavigationTest`.

### DIFF
--- a/src/aria/jsunit/RobotPhantomJS.js
+++ b/src/aria/jsunit/RobotPhantomJS.js
@@ -80,6 +80,7 @@ module.exports = Aria.classDefinition({
             }
         }
         keys.VK_CTRL = robotKeys.Control;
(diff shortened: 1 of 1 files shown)


simonarbuckle:
:ok: for me, you can integrate 65dd05d27cd1a63cb8857e0e223fdde9b2b91c70 for 1.7.17

divdavem:
@simonarbuckle Thank you for your review!

PR #1591, opened 2016-02-12
Adding the autoFocus property on the aria:Dialog widget

divdavem:
The new `autoFocus` property allows to control whether the first element in the dialog is automatically focused when the dialog is displayed.
By default, it is true for modal dialogs and false for other dialogs.

(cf #1577)

PR #1595, opened 2016-02-16
document.activeElement can be null on some browsers

divdavem:
This PR fixes an issue with commit ba8b45d (#1591): as `document.activeElement` can be null on some browsers, it is better to check whether it is non-null before using it.


simonarbuckle:
79d9f6f51b441e0fda8b8215e9c021245533ef55 :ok: for me

divdavem:
@simonarbuckle Thank you for your review

PR #1596, opened 2016-02-17
fix #1596 fixed the "Modal Dialog" test on IE 8 and made a few more robust

ymeine:
- now links are focused by clicking on them instead of using the programmatic way `.focus`
- all links used to get focus have now a `tabindex` property and no `href`, in order to avoid unwanted navigation or else
- made the "Dialog escape test" more robust
- fixed the "Modal Dialog" test on IE 8 by using the proper element's property


divdavem:
:ok: dad715d8935e0393cdf36eb1dd9dd7916d15033f seems fine for me

PR #1597, opened 2016-02-17
In aria.DomEvent, on Edge, isGecko was incorrectly true

divdavem:
This PR changes the value of the `isGecko` variable in `aria.DomEvent`, which was incorrectly true on Edge. This fixes several regressions introduced on Edge after changing `isSpecialKey` in commit e631660e271afe9352ddc36aa2c53873dc372a98.


ymeine:
Looks good, you can integrate it when the build passes

divdavem:
@ymeine Thank you for your review.

PR #1601, opened 2016-02-18
fix #1601 fixed the wai Modal dialog (test & implementation) and made a few more robust

ymeine:
- the biggest change is that I have split the wai Modal Dialog test to allow parallelization and avoid timeouts too (I therefore removed the timeout increase for PhantomJS)
- made some tests even more robust by adding a delay before clicks to avoid unwanted double-clicks
- fixed the Dialog implementation in IE7 (hiding elements): Element.hasAttribute doesn't exist
- fixed the wai Modal Dialog test in Firefox 3: Element.parentElement doesn't exist

closes #1601


divdavem:
:ok: 51185e98c722b6715791a5d7c5fb46d7f2021be0 seems fine for me.

PR #1602, opened 2016-02-18
fix #1602 The modal dialog JAWS test was not properly updated and failed

ymeine:


divdavem:
:ok: 7b5ff48add5c1b1ab8692a4f960564c1a2c4256a seems fine for me.

PR #1604, opened 2016-02-19
fix/test ClosePopupScrollTest on IE 9

divdavem:
This PR fixes `test.aria.widgets.container.dialog.closePopupScroll.ClosePopupScrollTest`
on IE9. It seems that auto-focusing the first element in the dialog prevents the following `scrollIntoView` on IE 9 (only).


benouat:
@divdavem :ok:

divdavem:

PR #1606, opened 2016-02-24
createATSyncContext method to use Aria Templates from node.js more easily

divdavem:
This PR adds the `createATSyncContext` method to the ariatemplates npm package.

Here is an example showing how it can be used (from node.js) to load the `aria.templates.TplClassGenerator` class and compile a template synchronously:

```js
var createATSyncContext = require("ariatemplates/src/node").createATSyncContext;
var context= createATSyncContext();
context.Aria.load({
    classes : ["aria.templates.TplClassGenerator"]
});
context.execTimeouts(); // executes any timeout synchronously
var classDef;
context.aria.templates.TplClassGenerator.parseTemplate("{Template {$classpath:'a.b.C'}} {macro main()}It works!{/macro}{/Template}", true, function(response) {
    classDef = response.classDef;
});
context.execTimeouts(); // executes any timeout synchronously
console.log(classDef); // here, classDef contains the generated class definition
```

An previous version of the `createATSyncContext` function is included in [atpackager](https://github.com/ariatemplates/atpackager/blob/v0.2.7/lib/ariatemplates.js), but it is probably better to share it in the Aria Templates package itself so that it can be used more easily by any project.

### DIFF
--- a/src/node/createATSyncContext.js
+++ b/src/node/createATSyncContext.js
@@ -1,0 +1,178 @@
+/*
+ * Copyright 2013 Amadeus s.a.s.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


fbasso:
9c09b89 is fine for me

divdavem:

PR #1607, opened 2016-02-25
Using Aria["eval"] instead of the eval statement in several places

divdavem:
This PR replaces the `eval` statement with the `Aria["eval"]` call so that the scope of the function calling `eval` is no longer exposed, and better minification can be achieved.


benouat:
@divdavem sounds :ok: to me

divdavem:

PR #1608, opened 2016-02-25
fix Enable focus binding in the error list widget

fbasso:
AT-1300


divdavem:
:ok: 346e6683d977aa4c2f810e3be17afbd5599c7f29 seems fine for me

PR #1609, opened 2016-03-04
fix WAI-ARIA issue with the aria:AutoComplete widget in Jaws 16

divdavem:
This PR fixes the following issue with the aria:AutoComplete widget in Jaws 16: when the suggestions list appears, the status was most of the time not read to notify the user that some suggestions are available.


ymeine:
Fine for me, you can integrate

divdavem:

PR #1610, opened 2016-03-04
Adding the codeMatch option in LCResourcesHandler

divdavem:
This PR adds the `codeMatch` option in `LCResourcesHandler` which defaults to `true` (previous behavior). When set to `false`, the `codeMatch` option disables matching on the code (to only match on the label).


ymeine:
Looks good to me, you can merge!

divdavem:

PR #1612, opened 2016-03-15
fix Invalid scope in dialog after a refresh

fbasso:
PTR 11293942, AT-1342


divdavem:
:ok: e01feee7f607f710c1fbac9d5d0c3288eb30b2b4 is fine for me.

PR #1613, opened 2016-03-17
fix checked/not checked on checkboxes wrongly reported by JAWS

fbasso:
AT-1310


divdavem:
:ok: 7a7476eb4f37887fd267d011657fe07c52a1c0c5 seems fine for me.
As discussed, the test will have to be added later.

PR #1615, opened 2016-03-24
fix Missing tests after commit fdafb39

fbasso:


divdavem:
:ok: e911125aafca8bcc620b52bbb8649b11e8f15235 seems fine for me.

PR #1616, opened 2016-03-25
Fixing DatePickerJawsTest1 with Jaws 16

divdavem:
It seems that Jaws 16 no longer say "entering application region" nor "leaving application region" in this test.


fbasso:
6eb928a is fine for me

divdavem:

PR #1622, opened 2016-04-05
feat New parameter added to load a test template without running the test

fbasso:
`norun="true"` can be now added to the test url.
Useful feature to debug or code on a template test case without having the full test to run, or the robot activated.


divdavem:
:ok: d404a83b5a68e998bb017e46918c6ece85599e8f seems fine for me.

PR #1623, opened 2016-04-07
fix Error list and error tooltips were not read with JAWS 16

fbasso:
AT-1352


divdavem:
:ok: 7d5884d585d3050778bb6100a69acdddaa09cd4a is fine for me

PR #1624, opened 2016-04-07
fix A global filter has been set for the JAWS tests

fbasso:


src/aria/jsunit/JawsTestCase.js
```diff
@@ -133,7 +133,9 @@ module.exports = Aria.classDefinition({
         assertJawsHistoryEquals : function (expectedOutput, callback, filterFn) {
             this.retrieveJawsHistory({
                 fn: function (response) {
+                    var globalFilterRegExp = /(Aria Templates tests frame|Aria Templates tests document)\n/gi;
 
+                    response = response.replace(globalFilterRegExp, "");
```
divdavem: Maybe I would put this filter in `retrieveJawsHistory` instead, as it is done for `"New tab page"`.

divdavem:
:ok: 9559b7fcc660febbd3a914e2d77fcf3e3e86c6dd seems fine for me

PR #1625, opened 2016-04-08
fix Render CheckboxJawsTestCase more stable

fbasso:
Because of random "not cheched/checked" or "checked/not checked" randomly read by JAWS


divdavem:
:ok: 8f25b3b0f72b3d647a1c9ebdd608913eae692b51 seems fine for me.

PR #1626, opened 2016-04-12
Bugfix/trim safer

MS-elug:


MS-elug:
Hello team,
That's my first Pull Request to your project.
Did I follow all the guidelines ?
Mathieu

ymeine:
Hello!

Regarding the form of the PR, the only thing to do is to squash all the commits together (it can be done before merging though).

Regarding the content then, here are my remarks: 

- the documentation of the method should be updated
- the test could be renamed (like `testSafeTrim`), and other values could be tested, since it's not related to `null` only but rather to non-String values
- coding style: add spaces around operators (after the `===`, between the `?` and the `:`)

Beyond that, I'm more concerned about the feature itself, for which I totally understand the motivation behind. However, this makes the contract change, so it's theoretically  non-backward compatible, even though it's actually a particular case: it's turning __errors__ to __silent bypassing of the feature__. This is something we might not want (what are your thoughts @ariatemplates/framework-core-contributors ?), and  would rather suggest adding an additional method, say `safeTrim`, doing the check and delegating to `trim` if appropriate. If you really want it to be the default, you could _patch_ the `aria.utils.String` class __carefully__ to replace `trim` with `safeTrim`.

divdavem:
Hello!

@MS-elug Thank you very much for your pull request. Your feedback is really appreciated!

To answer @ymeine's question, I would not call this change "non backward compatible" as the method is currently not supposed to be called with a non-string argument (when we extend the original scope of a method, such as adding an argument or accepting new types of arguments, these changes are usually not considered non-backward-compatible). I think opening this method to non-string arguments to pass them unmodified is useful. Even if it may turn some errors to silent, the result of the function is usually further processed (so errors could be raised there too if the result is not a string). I would not see the benefit of adding another method just for this additional check.
So, in my opinion, we could simply integrate this change...

divdavem:
It would be nice, though, to update the jsdoc documentation to explain that non-string parameters are returned unmodified.

ymeine:
Well, then some people might want the method to automatically convert any input to a String since the method expects to have a String instance. This is one example among others.

Applying `trim` to the input is the initial intent of the method, everything else is arbitrary, subject to variants, and suits different needs. Implementing one behavior or another is not an anodyne choice.

And I maintain that it could be non-backward compatible, but we didn't choose the error case to be part of the contract so it's not (and I agree with your explanation that it might eventually trigger an error during further processing).

PR #1627, opened 2016-04-12
Template parser and class generator improvements for tools

divdavem:
This PR improves the template parser and class generator in the following ways:
- the template tree now includes the processed source (which is the reference string for positions provided in the template tree)
- during class generation, most statements now add a "properties" property on their corresponding template tree node, to contain the parsed version of their parameters, so that it is possible to avoid code duplication with tools which analyse templates and need those pieces of information too
- the signature of the parseTemplate method in class generators was simplified so that it is easier to add options. The following options were added:
  - parseOnly: if true, the resulting class will not be generated, but the template is still fully parsed, including all statement properties
  - dontLoadWidgetLibs: If true, widget libraries referenced in the template will not be loaded during the class generation process, which is convenient if they are not available at that time. However, as a result, there will probably be missing dependencies in the generated class (but that is especially not a problem when parseOnly is true).

### DIFF
--- a/src/aria/templates/ClassGenerator.js
+++ b/src/aria/templates/ClassGenerator.js
@@ -113,16 +113,13 @@ module.exports = Aria.classDefinition({
          * parameter given to the callback function is: { classDef: {String} if null, errors occured during parsing or
          * class generation; otherwise contains the generated class }
          * @param {String} template the template
-         * @param {Boolean} allDeps If true, all dependencies should be included in the generated class. Otherwise only
-         * put classes which are not currently loaded in the $dependencies section.
+         * @param {aria.templates.CfgBeans:ClassGeneratorCfg} options Options for the class generation.
          * @param {aria.core.CfgBeans:Callback} callback the callback description
-         * @param {Object} context - passes template context information to improve debugging
-         * @param {Boolean} debug debug mode
          */
-        parseTemplate : function (template, allDeps, callback, context, debug, skipLogError) {
+        parseTemplate : function (template, options, callback) {
             var tree, errors;
             try {
-                tree = this._parser.parseTemplate(template, context, this.STATEMENTS, skipLogError);
+                tree = this._parser.parseTemplate(template, options.errorContext, this.STATEMENTS, options.skipLogError);
             } catch (e) {
                 errors = e.errors || e;
             }
@@ -133,20 +130,19 @@ module.exports = Aria.classDefinition({
                 });
                 return;
             }
-            this.__buildClass(tree, allDeps, callback, context, debug, skipLogError);
+            this.__buildClass(tree, options, callback);
         },
 
         /**
          * Parse a debug template from the tree
          * @param {aria.templates.TreeBeans:Root} tree tree returned by the parser.
-         * @param {Boolean} allDeps If true, all dependencies should be included in the generated class. Otherwise only
-         * put classes which are not currently loaded in the $dependencies section.
+         * @param {aria.templates.CfgBeans:ClassGeneratorCfg} options Options for class generation.
          * @param {aria.core.CfgBeans:Callback} callback callback the callback description
          * @param {Object} context - passes template context information to improve debugging
          * @param {Boolean} debug debug mode
          */
-        parseTemplateFromTree : function (tree, allDeps, callback, context, debug, skipLogError) {
-            this.__buildClass(tree, allDeps, callback, context, debug, skipLogError);
+        parseTemplateFromTree : function (tree, options, callback) {
+            this.__buildClass(tree, options, callback);
         },
 
         /**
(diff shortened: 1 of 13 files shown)


src/aria/templates/ClassGenerator.js
```diff
@@ -133,20 +136,25 @@ module.exports = Aria.classDefinition({
...
 
         /**
          * Parse a debug template from the tree
          * @param {aria.templates.TreeBeans:Root} tree tree returned by the parser.
-         * @param {Boolean} allDeps If true, all dependencies should be included in the generated class. Otherwise only
-         * put classes which are not currently loaded in the $dependencies section.
+         * @param {aria.templates.CfgBeans:ClassGeneratorCfg} options Options for class generation.
          * @param {aria.core.CfgBeans:Callback} callback callback the callback description
          * @param {Object} context - passes template context information to improve debugging
          * @param {Boolean} debug debug mode
```
ymeine: I think you forgot to remove documentation about the deprecated parameters `context` and `debug`.

src/aria/templates/ClassGenerator.js
```diff
@@ -133,20 +136,25 @@ module.exports = Aria.classDefinition({
...
          */
-        parseTemplateFromTree : function (tree, allDeps, callback, context, debug, skipLogError) {
-            this.__buildClass(tree, allDeps, callback, context, debug, skipLogError);
+        parseTemplateFromTree : function (tree, optionsOrAllDeps, callback, context, debug, skipLogError) {
+            var options = (typeof optionsOrAllDeps == "object") ? optionsOrAllDeps : {
+                allDependencies: optionsOrAllDeps,
+                errorContext: context,
+                debug: debug,
+                skipLogError: skipLogError
+            };
```
ymeine: This code is duplicated from method `parseTemplate`. I think it could be made common in a private method.

src/aria/templates/ClassGenerator.js
```diff
@@ -113,16 +113,19 @@ module.exports = Aria.classDefinition({
...
             var tree, errors;
             try {
-                tree = this._parser.parseTemplate(template, context, this.STATEMENTS, skipLogError);
+                tree = this._parser.parseTemplate(template, options.errorContext, this.STATEMENTS, options.skipLogError);
```
ymeine: Since you use the `options` here you should maybe do the same thing as in `__buildClass`: validating the object. I understand that then there would be more work to avoid duplicate validation since this `__buildClass` method is called afterwards.

src/aria/templates/ClassWriter.js
```diff
@@ -495,6 +507,15 @@ module.exports = Aria.classDefinition({
...
             }
+        },
+
+        /**
+         * Set the parseOnly flag to true and disables functions which are only useful for class generation.
+         */
+        enableParseOnly : function () {
+            this.parseOnly = true;
+            // disables functions which are only useful to generate code:
+            this.writeDependencies = this.newVarName = this.writeln = this.write = this.wrapExpression = this.trackLine = Aria.empty;
```
ymeine: I would rather split this line (even though it duplicates code): 

```javascript
this.writeDependencies = Aria.empty;
this.newVarName = Aria.empty;
this.writeln = Aria.empty;
this.write = Aria.empty;
this.wrapExpression = Aria.empty;
this.trackLine = Aria.empty;
```

With nowadays editors features, it's easy to add/remove a method (duplicate/delete line), and visually it's perfectly aligned so that we can know more easily the set of overridden methods at a glance.

src/aria/templates/Statements.js
```diff
@@ -142,7 +142,7 @@ module.exports = Aria.classDefinition({
                     parts.push(param);
 
                     // Automatic escape ----------------------------------------
-
+                    var autoEscape = false;
```
ymeine: I would just add a blank line above, to keep (consistent) spacing as before.

src/aria/templates/Statements.js
```diff
@@ -162,13 +162,14 @@ module.exports = Aria.classDefinition({
...
                                 parts.push(escapeModifierName);
+                                autoEscape = true;
                             }
                         }
 
                     }
 
                     // end of: automatic escape --------------------------------
-
+                    var modifiersList = [];
```
ymeine: Same as above ;)

test/aria/templates/TplClassGeneratorParseOnlyTest.js
```diff
@@ -0,0 +1,101 @@
...
+                "args": '"dd/MM/yyyy"'
+            }]);
+
+            var widgetStatement = testMacroStatement.content[2];
+            this.assertEquals(widgetStatement.name, "@myLib:MyWidget");
+            this.assertEquals(widgetStatement.properties.libName, "myLib");
+            this.assertEquals(widgetStatement.properties.libClasspath, "x.y.MyWidgetsLibThatDoesNotExistButShouldNotBeChecked");
+            this.assertEquals(widgetStatement.properties.widgetName, "MyWidget");
+
+            this.assertTruthy(result, "The callback should have been called synchronously.");
```
ymeine: Shouldn't you test this first? I (only) now understand why you were assuming the `result` object was immediately available.

ymeine:
This is good for me, you can integrate it. Btw, these are awesome features :wink:

divdavem:
@ymeine Thank you for your kind review!

PR #1628, opened 2016-04-14
fix Fixed the TabsJawsTest (by filtering) and improved the JAWS test utilities

ymeine:
- the `TabJawsTest` now ignores irrelevant content, especially the links that are clicked to have focus.
- the `EnhancedJawsTestCase` now supports the use of a filter
- the `JawsTestCase` now logs more information when there is an assert error


divdavem:
:ok: a7d119e6145ce7cf93191612e5e4605687217eed seems fine for me

PR #1633, opened 2016-04-21
WAI-ARIA adding support for aria label related attributes in the link widget

simonarbuckle:
**Adds 3 new properties for the link widget:**

- `waiLabel:` sets 'aria-label' on the link, value will be used as the attributes value
- `waiDescribedBy:` sets 'aria-describedby' on the link, value will be used as the attributes value
- `waiLabelledBy:` sets 'aria-labelledby' on the link, value will be used as the attributes value

_Note: the 'aria-hidden' property is not needed as there is no label element for the link widget._


divdavem:
:ok: d6e7918e499d7c544ff655f0dc3ec6a85315b884 seems fine for me.

PR #1634, opened 2016-04-26
feat WAI-ARIA fixes an issue with the TextField widget when it is a required field

simonarbuckle:
The attribute `aria-required="true"`is now added to the widgets input dom element when the widget property `mandatory: true` is set.


divdavem:
:ok: ec31356e4f91c5ea12f631ab562e5e35e0b6d17c seems fine for me.

PR #1635, opened 2016-04-28
feat WAI-ARIA fixes an issue with the Fieldset widget label being read at the end of the Fieldset

simonarbuckle:
This change marks the current label with "aria-hidden", and adds another span containing the class "xSROnly" with the label content, and renders this span in the correct position before the frame.

_Note: this change is only switched on when accessibility is enabled._


divdavem:
:ok: 6095967264cc3eb44c7ee540a394209703ec7a51 seems fine for me.

PR #1636, opened 2016-04-29
feat WAI-ARIA fixes an issue with the SelectBox where the list of results is not announced by JAWS

simonarbuckle:
In the SelectBox, when there are suggestions Jaws doesn't announce that there are suggestions.  Should work in the same way as the AutoComplete, for example if there are 3 items in the list Jaws should say "There are 3 suggestions".


divdavem:
:ok: 1933e8d8f725df5ef704e4d720e2890f72a34f2e seems fine for me

PR #1638, opened 2016-05-03
Injecting DomEventWrapper in TextField:onclick/onfocus/onblur

benouat:
As the name suggest, we were simply missing some events to be injected in their proper callbacks execution.
This PR fixes it for `onclick`, `onfocus` and `onblur`. It impacts all widgets that inherit from `aria.widgets.form.TextInput`.

I only tested on TextField and SelectBox. I think it's enough.


benouat:
@divdavem @simonarbuckle anyone of you two could have a look at this one ? thanks

src/aria/widgets/form/TextInput.js
```diff
@@ -910,9 +912,13 @@ module.exports = Aria.classDefinition({
...
          */
-        _dom_onclick : function () {
+        _dom_onclick : function (event) {
             if (!!this._cfg.onclick) {
-                this.evalCallback(this._cfg.onclick);
+                var evtWrapper = new ariaTemplatesDomEventWrapper(event);
+                this.evalCallback(this._cfg.onclick, evtWrapper);
+                if (evtWrapper) {
+                    evtWrapper.$dispose();
+                }
```
divdavem: I don't see how `evtWrapper` could be falsy here... so I would remove the condition here (and also in the next two cases) and call `evtWrapper.$dispose()` unconditionally.

divdavem:
When searching for `_dom_onblur`, `_dom_onfocus` and `_dom_onclick` in the full source code of Aria Templates, it seems that there are some cases where those methods are called internally (especially in the DatePicker) without any event, which seems to raise an exception when the constructor of DomEventWrapper is called (and its parameter is undefined).

PR #1642, opened 2016-06-01
Chrome keeps the selected option after clearing options in the `@aria:Select`

divdavem:
This PR fixes a strange issue in Chrome: when removing all options from an `@aria:Select` widget, Chrome sometimes keeps the selected option displayed. Note that this does not always happen, it depends on which option is selected.


fbasso:
c69ee2a is fine for me

divdavem:

PR #1643, opened 2016-06-09
fix #1643 The AutoComplete's dropdown was sometimes closed too early because of unexpected blur events preceding others

ymeine:
The issue was introduced in commit: f123c7a173b77fcd627f68089da702c8b57fc2bf released in version 1.7.15

Under some execution environment, it happened that some blur events were fired on the AutoComplete's input element before any other event were. The initial commit was making the widget rely on blur to close the its dropdown, which in this unfortunate case prevented other events to occur on this same dropdown, that way preventing item selection from the mouse for instance.

PTR 11565758


divdavem:
:ok: e22f29672389f2dee94f751b7e7ef9dd83ffc232 seems fine for me.

divdavem:
:ok: 95a1b2bea287cc570e23edc5c054593df42ad6e5 seems fine for me (if the travis build is ok).

PR #1644, opened 2016-06-23
Listening to tap events for popup icons of DatePicker and similar widgets

divdavem:
In order to fix issues when trying to open the DatePicker popup (or popups of similar widgets), we now listen to the tap event instead of the click event on touch devices.

Touch devices are detected using `aria.utils.Device.isTouch()`.
`aria.touch.Tap` is only loaded if a touch device is detected.

### DIFF
--- a/src/aria/touch/Event.js
+++ b/src/aria/touch/Event.js
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 var Aria = require("../Aria");
-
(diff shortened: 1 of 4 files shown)


benouat:
@divdavem I am fine with the change. I am still not yet familiar with dynamic noder modules :) but that's ok :P

divdavem:

PR #1646, opened 2016-07-01
Making dialog title and buttons focusable or not by configuration

divdavem:
This PR adds 3 new configuration options for `aria:Dialog` widgets:
- `focusableTitle`: whether the title of the dialog is focusable. By default, it is set to true if `waiAria` is true.
- `focusableClose` and `focusableMaximize`: whether the corresponding button is focusable. By default, it is set to true if `waiAria` is true and the corresponding label is defined.


benouat:
@divdavem I am fine with these changes.

divdavem:

PR #1650, opened 2016-07-06
Adding the titleTag property on the Dialog widget

divdavem:
This PR adds the `titleTag` property to the Dialog widget to allow setting the tag used for the title in the dialog.
By default, the tag used for dialogs is now `h1` when `waiAria` is `true` and `span` (as before) otherwise.

This is especially useful for accessibility: screen readers often provide a way to navigate through the heading tags of the page.


fbasso:
5c89ec6 is fine for me

divdavem:

PR #1651, opened 2016-07-06
Adding the waiEscapeMsg and waiCloseMsg properties on the Dialog widget

divdavem:
This PR adds the following properties on the dialog widget:

- `waiEscapeMsg`: If this property is defined and if `waiAria` is activated, the user has to press escape twice (instead of only once) to close the dialog. This property specifies the message to read after the user pressed escape the first time.

- `waiCloseMsg`: When `waiAria` is activated, this property specifies the message to read when the modal dialog is closed.


fbasso:
cefa54a is fine for me

divdavem:

PR #1653, opened 2016-07-07
feat Add commented options in attester

fbasso:
Useful to just un-comment these options when required in dev mode


divdavem:
:ok: cbf1e16890b62ca780126aab0f7e44bc3921ce65 seems fine for me.

PR #1654, opened 2016-07-07
feat #1654 WAI-ARIA - Modal dialog - Focus cycling for accessible modal popups

ymeine:
Now the focus is maintained inside the popup when navigating with the tab key.

This will loop through the elements of the popup, taking into account forward and backward navigation.

This will also handle properly the focus and the navigation direction when navigating from the user agent GUI.

PTR 11592232
AT-1425
AT-1442
AT-1455


benouat:
@ymeine After the review, I am fine with this PR.
You can merge it as soon as travis is green 👍

PR #1655, opened 2016-07-07
Jaws reads "not checked unavailable" when the checkbox is checked (and disabled)

divdavem:
This PR fixes the following issue in IE 11 / Jaws 16: in some cases a disabled checked checkbox is read as "not checked unavailable" instead of "checked unavailable".


benouat:
@divdavem these changes are fine for me. you can merge it

divdavem:
@benouat Thank you for your review! :smile:

PR #1659, opened 2016-07-27
Initially disabled radio buttons misbehaving when enabled

divdavem:
Initially disabled radio buttons were excluded from the array of all radio button instances and never added (even when re-enabled), which prevented them from reacting correctly to arrow up and down navigation.

This commit makes sure all radio buttons (whether they are enabled or not) are present in the array of all radio button instances. Disabled radio buttons are simply skipped when looping through the array.


benouat:
@divdavem  theses changes seems fine. You can merge.
If you want to provide one more test, or change this one to be a standard robot one instead of a Jaws one... Feel free !

divdavem:
@benouat Thank you for your review. I will add a standard robot test.

PR #1662, opened 2016-07-29
Disabled button to actually be disabled [JAWS]

benouat:
No matter the button is a simpleHtml one or not. we always set the button as disabled via the proper attribute when needed.
We previously were setting it only in case of a simpleHTML button.

I am doing it all the time, there is no need is to check that `waiAria` is active.

### DIFF
--- a/src/aria/widgets/action/Button.js
+++ b/src/aria/widgets/action/Button.js
@@ -149,13 +149,7 @@ module.exports = Aria.classDefinition({
             if (!skipChangeState) {
                 // force widget - DOM mapping
                 this.getDom();
-                if (this._simpleHTML) {
-                    if (state == "disabled") {
-                        this._focusElt.setAttribute("disabled", "disabled");
-                    } else {
-                        this._focusElt.removeAttribute("disabled");
-                    }
-                } else {
+                if (!this._simpleHTML) {
                     this._frame.changeState(this._state);
                     var ie8plus = ariaCoreBrowser.isOldIE && ariaCoreBrowser.majorVersion >= 8;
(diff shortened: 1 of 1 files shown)


benouat:
We just need to double check that everything is fine on IE7 where we don't have a button.
It should be fine, but did not yet had time to test it.

benouat:
@divdavem @fbasso All tests are green. Do you think we could merge it ?

src/aria/widgets/action/Button.js
```diff
@@ -149,29 +149,24 @@ module.exports = Aria.classDefinition({
             if (!skipChangeState) {
                 // force widget - DOM mapping
                 this.getDom();
-                if (this._simpleHTML) {
-                    if (state == "disabled") {
-                        this._focusElt.setAttribute("disabled", "disabled");
-                    } else {
-                        this._focusElt.removeAttribute("disabled");
+                this._frame.changeState(this._state);
```
divdavem: There is no frame when `this.simpleHtml` is true, so this line will raise an exception.
It is strange that unit tests did not catch this problem!

divdavem:
`_widgetMarkup` should also be modified so that if the button is disabled, the initial markup contains the `disabled` attribute.

benouat:
@divdavem I somehow reverted my change to keep in mind that when simpleHTML is `true` there is no frame. I also modified the `_widgetMarkup()` method accordingly.

divdavem:
:ok: c6bb070f9f086007bd4fc33f313ef168d7db3832 seems fine for me.

PR #1664, opened 2016-08-18
Fixed incorrect conversion of 'mousewheel' event for Chrome when removing an event listener

vcarle:
Hello,

I have noticed an inconsitency in the logic to handle the **mousewheel** event between the **addListener** and the **removeListener** methods of the **aria.utils.Event** class.

For Chrome, the event is converted to a **DOMMouseScroll** in the **removeListener** but not in the **addListener**. As a consequence, **mousewheel** event listeners are never removed, which can lead to memory leaks.

As a fix. I am now making sure the **mousewheel** event conversion to **DOMMouseScroll** does not apply to Chrome.

IMPORTANT: there is another discrepancy between the 2 methods:

- **addListener** checks **isIE**
- **removeListener** checks **isOldIE**

I do not know which one should be kept, so I have not modified this logic, but fore sure the 2 methods should be aligned (and this bit of code factorised ;)).

Best regards,
Vincent.


divdavem:
Hello Vincent,
Thank you for reporting this issue and opening a pull request.
As you say, for the two methods to be aligned, it would be good to factorize this code to be sure we always have the same logic applied in `addListener` and `removeListener`.
Would you like to do it?

Regarding the other discrepency you mention:
`isOldIE` and `isIE` always have the same value, except on IE 11 where `isOldIE` is `false` and `isIE` is `true`.
Unless we specifically test on IE 11 what is the right event to use, we should keep in the factorized function the logic from `addListener` (rather than `removeListener`), so that the behavior of `addListener` does not change, so it would be `isIE` rather than `isOldIE`

vcarle:
PR updated :)

divdavem:
@vcarle Thank you! :smile:

PR #1666, opened 2016-08-19
feat #1666 WAI-ARIA - ActionWidget - Implementation of labels, description and disabled

ymeine:
The impacted widgets are the following:

- ActionWidget
	- Button, including the simple button where the element is now a "button" instead of an "input"
		- IconButton
	- Link
	- SortIndicator

Implemented attributes are (on the left the WAI-ARIA attribute name, on the right the widget's property changing its value):

- aria-label: waiLabel
- aria-labelledby: waiLabelledBy
- aria-describedby: waiDescribedBy
- aria-disabled: disabled

### DIFF
--- a/src/aria/utils/String.js
+++ b/src/aria/utils/String.js
@@ -119,7 +119,7 @@ module.exports = Aria.classDefinition({
          *
          * @param {String} str Input string
          *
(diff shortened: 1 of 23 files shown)


divdavem:
Please fix the undisposed object error (about the view).

PR #1667, opened 2016-08-23
Improving the readability of Jaws test case failures

divdavem:
This PR adds some line breaks in the error message of Jaws test cases so that it is easier to compare the actual Jaws history with the expected one.


ymeine:
Thanks, this helps a lot! You can integrate it!

divdavem:

PR #1668, opened 2016-08-23
Adding support for SIGINT for "npm run attest"

divdavem:
This PR allows to use CTRL-C when running the "npm run attest" command, in order to interrupt it properly (including on Windows).

Cf the similar commit in attester: https://github.com/attester/attester/commit/478634091e9e8abe80dad9aae12a230ac432b5d1


fbasso:
094e26d is fine for me

divdavem:

PR #1669, opened 2016-08-24
fix wrong flag in filtering regular expression of JawsTestCase

divdavem:


divdavem:
Now that #1670 is integrated, this PR is no longer useful.

PR #1670, opened 2016-08-24
Factorizing filters for Jaws test cases, improves the stability of those tests

divdavem:
This PR refactors the way filters for Jaws history are applied in Jaws test cases:
* `retrieveJawsHistory` no longer does any filtering, it returns the raw Jaws history
* `assertJawsHistoryEquals` now first executes `filterJawsHistory` before any provided filter function
* `filterJawsHistory` executes by default the 3 following common filters on all tests:
   * `normalizeSpaces` (can be disabled with `this.skipNormalizeSpaces = true`)
   * `removeNoise` (can be disabled with `this.skipRemoveNoise = true`)
   * `removeDuplicates` (can be disabled with `this.skipRemoveDuplicates = true`)
* `removeNoise` can be configured by changing the array of regular expressions stored in `this.noiseRegExps`

### DIFF
--- a/src/aria/jsunit/JawsTestCase.js
+++ b/src/aria/jsunit/JawsTestCase.js
@@ -14,6 +14,7 @@
  */
 var Aria = require("../Aria");
 var ariaUtilsString = require("../utils/String");
+var ariaUtilsArray = require("../utils/Array");
 
 /**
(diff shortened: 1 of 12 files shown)


ymeine:
This looks good to me!

divdavem:

PR #1673, opened 2016-08-29
fix/test #1673 test.aria.widgets.wai.errorlist.binding.ErrorListBindingJawsTestCase had wrong expected history

ymeine:
The expected history was not correct.


divdavem:
:ok: 095fcdffeacd96e9d850e23596af9b4ac92e5245 seems fine for me.

PR #1674, opened 2016-08-30
fix aria-activedescendant initially not set for the autocompletete widget

divdavem:
When using the `@aria:AutoComplete` component with `waiAria = true`, the `aria-activedescendant` attribute was not set until the user navigates the suggestions with up/down arrow keys.

This was an issue when a suggestion is preselected while the user types (which can happen, for example, if the `preselect` property is set to `"always"`)


fbasso:
82b275a is fine for me

divdavem:

PR #1675, opened 2016-08-31
feat #1675 Implement the "disabled" attribute for the Link widget

ymeine:
Also made the following change to all "ActionWidget" based widgets: now the "aria-disabled" attribute is not set anymore in WAI-ARIA mode, "disabled" must be sufficient.

For now, this PR depends on https://github.com/ariatemplates/ariatemplates/pull/1662 Without that, the following tests are failing (even though the final expected output has already been fixed):

- test.aria.widgets.wai.input.actionWidget.Button
- test.aria.widgets.wai.input.actionWidget.IconButton

References:

- AT-1398
- PTR 11630074


divdavem:
:ok: 6d58cab1316fe809ad5e2249307df8b562c60a61 seems fine for me.

Now that #1662 is integrated, this PR can be integrated.

You probably planned to do it, but you should not forget to remove the comment about the dependency on #1662 and the failing tests from the commit message before integrating the commit in master.

PR #1676, opened 2016-09-06
Configurable default values for any widget property of the Aria library

divdavem:
This PR adds the `widgetDefaults` parameter in the environment, allowing to specify a default value for any property of any widget in the Aria widget library.

For example, to set the `calendarNumberOfUnits` property for all instances of the `@aria:DatePicker` widget (if those instances do not specify any value for this property):

```js
  aria.core.AppEnvironment.setEnvironment({
    widgetDefaults: {
      DatePicker: {
        calendarNumberOfUnits: 5
      }
    }
  });
```


benouat:
@divdavem you can merge as soon as CI is green

divdavem:

PR #1680, opened 2016-09-09
Changing the names of tests

divdavem:
This PR changes the names of tests to follow the following convention:
- tests to be run with Jaws enabled should have a classpath ending with `JawsTestCase`
- tests to be run with the robot should have a classpath ending with `RobotTestCase`
- other tests should have a classpath ending with `TestCase`
- intermediate classes which both extend directly or indirectly `aria.jsunit.TestCase` and have child test classes should not be run directly as tests and should have a classpath ending with `JawsBase`, `RobotBase` or `Base` (depending on the case)

This will allow in the future to completely remove test suites and rely on naming conventions to run tests.

This pull request contains several commit to distinguish between two different automatic operations (renaming files and updating their content) and other manual operations.


benouat:
@divdavem I am fine with this huge PR. Nonetheless, I see you have split it nicely in plenty of commits. This is good. I would then suggest that we don't squash this PR, and we apply all commits instead. 
As we are always fast forward, simply push your branch to master as soon as everyone in the team is ware of that.

divdavem:

PR #1681, opened 2016-09-13
Remove test suites

divdavem:
Following #1680, this PR:
* changes test campaigns to use patterns instead of references to test suites
* removes test suites (which are now useless)


benouat:
:+1: !

PR #1682, opened 2016-09-14
New readOnlyTextField property on the aria:MultiSelect widget

divdavem:
This commit adds the `readOnlyTextField` property to the multiselect. If true, the user can only modify the value of the widget by opening the popup (the text field is not modifiable).


divdavem:

PR #1685, opened 2016-09-15
Updating Jaws tests after adding a default value for aria-labelledby

divdavem:
After commit 24960cb98cef30f45e0cd897c64e0c317eecddff, the following tests failed because the corresponding radio buttons now have a label (which was not expected in Jaws history):
- `test.aria.widgets.wai.fieldset.FieldsetLabelJawsTestCase`
- `test.aria.widgets.wai.input.radiobutton.RadioButtonGroupJawsTestCase`

This PR updates the expected history in those tests.


divdavem:

PR #1686, opened 2016-09-16
Add Array.prototype.every to aria.utils.Array. New test added to ArrayTestCase

tjw02:
This pull request would like to add the Array.prototype.every method to the aria.utils.Array object. This is to polyfill older versions of the browser and also to have a clean way of performing tasks on Arrays. The purpose of this addition is to allow us to write recursion that will replicate the break functionality, so we don't have to waste time looping through the list when the required item is found.

### DIFF
--- a/src/aria/utils/Array.js
+++ b/src/aria/utils/Array.js
@@ -182,6 +182,7 @@ var Aria = require("../Aria");
              * @param {Function} Function to test each element of the array. This function receives as arguments the
              * value, the index and the array.
              * @param {Object} thisObject Object to use as this when executing callback.
+             * @return {Boolean} true if every element in the given array returns truthy values for the predicate (callback).
              */
             every : (!arrayPrototype.every) ? function(array, callback, thisObject) {
                 // clone array to avoid mutation when executing the callback
--- a/src/aria/utils/Array.js
+++ b/src/aria/utils/Array.js
@@ -184,7 +184,7 @@ var Aria = require("../Aria");
              * @param {Object}   thisObject - Object to use as this when executing callback. The default value, when thisObject 
              *                                is falsy, will be the given array.
              * @return {Boolean} true if every element in the given array returns truthy values for the predicate (callback) 
-             *                   false if at least one element returns a falsy value for the predicate (callback)
+             *                   false if at least one element returns a falsy value for the predicate (callback) 
              */
             every : (!arrayPrototype.every) ? function(array, callback, thisObject) {
                 // clone array to prevent being impacted by possible mutations of the array in the callback
--- a/src/aria/utils/Array.js
+++ b/src/aria/utils/Array.js
@@ -183,8 +183,8 @@ var Aria = require("../Aria");
              *                                value, the index and the array.
              * @param {Object}   thisObject - Object to use as this when executing callback. The default value, when thisObject 
              *                                is falsy, will be the given array.
-             * @return {Boolean} true if every element in the given array returns truthy values for the predicate (callback). 
-             *                   false if at least one element returns a falsy value for the predicate (callback).
+             * @return {Boolean} true if every element in the given array returns truthy values for the predicate (callback) 
+             *                   false if at least one element returns a falsy value for the predicate (callback)
              */
             every : (!arrayPrototype.every) ? function(array, callback, thisObject) {
                 // clone array to prevent being impacted by possible mutations of the array in the callback
@@ -193,7 +193,7 @@ var Aria = require("../Aria");
                 thisObject = thisObject || array;
                 var arrayLength = workArray.length;
                 for (var arrayIndex = 0; arrayIndex < arrayLength; arrayIndex++) {
-                    if(!callback.call(thisObject, workArray[arrayIndex], arrayIndex, array)) {
+                    if ( !callback.call(thisObject, workArray[arrayIndex], arrayIndex, array) ) {
                         return false;
                     }
                 }
--- a/src/aria/utils/Array.js
+++ b/src/aria/utils/Array.js
@@ -173,6 +173,30 @@ var Aria = require("../Aria");
                     }
                 }
                 return output;
+            },
+            /**
+             * This tests whether all elements in the array pass the test implemented by the provided function. 
+             * As long as the elements return truthy values, the loop will continue. However, as soon as a falsy
+             * value is returns, the answer will be returned without analyzed the rest of the elements.
+             * @param {Array} array Array to filter.
+             * @param {Function} Function to test each element of the array. This function receive as arguments the
+             * value, the index and the array.
+             * @param {Object} thisObject Object to use as this when executing callback.
+             */
+            every : (!arrayPrototype.every) ? function(array, callback, thisObject) {
+                // clone array to avoid mutation when executing the callback
+                var workArray = arrayUtils.clone(array);
+
+                thisObject = thisObject || array;
+                var len = workArray.length;
+                for (var i = 0; i < len; i++) {
+                    if(!callback.call(thisObject, workArray[i], i, array)) {
+                        return false;
+                    }
+                }
+                return true;
+            } : function(array, callback, thisObject) {
+                return array.every(callback, thisObject);
             }
         }
     });
--- a/src/aria/utils/Array.js
+++ b/src/aria/utils/Array.js
@@ -175,11 +175,11 @@ var Aria = require("../Aria");
                 return output;
             },
             /**
-             * This tests whether all elements in the array pass the test implemented by the provided function. 
-             * As long as the elements return truthy values, the loop will continue. However, as soon as a falsy
-             * value is returns, the answer will be returned without analyzed the rest of the elements.
-             * @param {Array} array Array to filter.
-             * @param {Function} Function to test each element of the array. This function receive as arguments the
+             * This tests whether all elements in the array pass the test given by the predicate callback function. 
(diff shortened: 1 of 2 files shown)


src/aria/utils/Array.js
```diff
@@ -173,6 +173,30 @@ var Aria = require("../Aria");
...
+             * value, the index and the array.
+             * @param {Object} thisObject Object to use as this when executing callback.
+             */
+            every : (!arrayPrototype.every) ? function(array, callback, thisObject) {
+                // clone array to avoid mutation when executing the callback
+                var workArray = arrayUtils.clone(array);
+
+                thisObject = thisObject || array;
+                var len = workArray.length;
+                for (var i = 0; i < len; i++) {
```
ymeine: Please use longer, meaningful variable names (and preferably full words): 

- `len` :right_arrow: `length`
- `i` :right_arrow: `index`

Also, technically you can define the variable `len`/`length` in the `for` statement, as for `i`/`index`. But both are perfectly fine.

src/aria/utils/Array.js
```diff
@@ -173,6 +173,30 @@ var Aria = require("../Aria");
...
+             * @param {Array} array Array to filter.
+             * @param {Function} Function to test each element of the array. This function receive as arguments the
+             * value, the index and the array.
+             * @param {Object} thisObject Object to use as this when executing callback.
+             */
+            every : (!arrayPrototype.every) ? function(array, callback, thisObject) {
+                // clone array to avoid mutation when executing the callback
+                var workArray = arrayUtils.clone(array);
+
+                thisObject = thisObject || array;
```
ymeine: Be careful that with this common shortcut for setting default values, you are preventing any falsy values to be used, even though they might be used in practice: `false`, `0`, `''` are all valid values for `thisArg`. Also, if the user wants to receive the global object as `thisArg`, he can't this way since he would have to specify `undefined` or `null`, which are both falsy values too.

Since the specifications http://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.every and the recommended polyfill implementation in MDN https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every#Polyfill don't even match on that point, I don't know what to decide about this.

My general opinion is to leave it as flexible as possible, and let the code fail by itself later on (while calling `callback.call` in this specific case) if the value was not correct, in order to respect the DRY concept and to rely as most as possible on standard implementations.

src/aria/utils/Array.js
```diff
@@ -173,6 +173,30 @@ var Aria = require("../Aria");
...
                 }
                 return output;
+            },
+            /**
+             * This tests whether all elements in the array pass the test implemented by the provided function. 
+             * As long as the elements return truthy values, the loop will continue. However, as soon as a falsy
+             * value is returns, the answer will be returned without analyzed the rest of the elements.
+             * @param {Array} array Array to filter.
+             * @param {Function} Function to test each element of the array. This function receive as arguments the
+             * value, the index and the array.
```
ymeine: Please precise that the received array is a __copy__.

src/aria/utils/Array.js
```diff
@@ -173,6 +173,30 @@ var Aria = require("../Aria");
...
+             * @param {Object} thisObject Object to use as this when executing callback.
```
tjw02: Hi, I took the name thisObject from the rest of the methods in this utils class, would you suggest renaming all of the thisObjects (for the filter and forEach methods) to thisArgument?

src/aria/utils/Array.js
tjw02: This line was taken also from the forEach and filter methods in this class. Would you like me to change all three, or just change this one so the behaviour is different to those two other methods?

src/aria/utils/Array.js
tjw02: Are you sure that it is a copy? It looks as if the original array is being passed through?

tjw02:
Hi,

I am in the processing of making the changes as requested, but I had some questions about a couple in relation to the filter and forEach methods in this Utils Object which the above comments would also apply to.

Could you decide what you would like the outcome to be for each of my questions?
Cheers

src/aria/utils/Array.js
```diff
@@ -173,6 +173,31 @@ var Aria = require("../Aria");
                     }
                 }
                 return output;
+            },
+            /**
+             * This tests whether all elements in the array pass the test given by the predicate callback function. 
+             * As long as the predicate (callback) returns truthy values, the loop will continue. However, as soon as a falsy
+             * value is returned, the function will return without analyzing the rest of the elements.
+             * @param {Array} array Array to analyze.
```
ymeine: Just a remark: `analyzing` and `analyze` are maybe a bit specific words: the array and its items can be ___processed___ without being _analyzed_.

src/aria/utils/Array.js
```diff
@@ -173,6 +173,31 @@ var Aria = require("../Aria");
...
+             * @param {Function} Function to test each element of the array. This function receives as arguments the
+             * value, the index and the array.
+             * @param {Object} thisObject Object to use as this when executing callback.
```
ymeine: Please precise what is the default value for `thisObject` and when it is applied (e.g. `thisObject` will be set to `array` when falsy)

src/aria/utils/Array.js
```diff
@@ -173,6 +173,31 @@ var Aria = require("../Aria");
...
+             * @return {Boolean} true if every element in the given array returns truthy values for the predicate (callback).
```
ymeine: Please precise the return value in the counterpart case (`false`)

src/aria/utils/Array.js
ymeine: Actually — after discussing with some members of the team — we think that it's better to keep consistency with the other methods (sorry I didn't pay attention to the rest of the module while reviewing) and to leave it like you wrote. Anyway the user will still be able to use any other value for `this` if he actually wants to.

Regarding the _mutation prevention_, sorry it was part of my confusion, this doesn't change anything.

src/aria/utils/Array.js
```diff
@@ -173,6 +173,31 @@ var Aria = require("../Aria");
...
+             */
+            every : (!arrayPrototype.every) ? function(array, callback, thisObject) {
+                // clone array to avoid mutation when executing the callback
```
ymeine: This comment — which I know you copied, no blame on you — actually got me confused. I looked at the code too fast then, and thought that this array was passed as the third parameter to the `callback`. Hence my comments for the documentation and for the setting of the default value for `thisObject` which are actually wrong.

I would suggest being a bit more precise and say something more like `to prevent being impacted by possible mutations of the array in the callback`.

test/aria/utils/ArrayTestCase.js
```diff
@@ -237,6 +237,46 @@ Aria.classDefinition({
...
+            var every = aria.utils.Array.every;
+
+            // doing stuff on empty array should do nothing
+            var count = 0;
+            every([], function () {
+                count++;
+            });
+            this.assertTrue(count === 0);
+
+            // forEach simple test
```
ymeine: `forEach` should be `every`, oversight after copy-paste always happens ;)

Anyway I think it's better to leave it as `simple test`, I don't see the point of repeating the name of the tested function here, it doesn't bring more information (and again, it's annoying when copy-pasting). My comment applies only to your code, no need to change the rest of the module.

test/aria/utils/ArrayTestCase.js
```diff
@@ -237,6 +237,46 @@ Aria.classDefinition({
...
+                // changing the array should not break
+                testArray = [];
+                result += value;
+                return value !== 4;
+            }, this);
+            this.assertTrue(result === "1234");
+
+            testArray = [1, 2, 3, 4, 5, 9], result = "";
+            every(testArray, function (value, index, array) {
+                // changing the array should not break
```
ymeine: No need to test that again (same thing for the next occurrence).

test/aria/utils/ArrayTestCase.js
```diff
@@ -237,6 +237,46 @@ Aria.classDefinition({
                 result += value;
             }, this);
             this.assertTrue(result === "123459");
+        },
+
+        testEvery : function () {
```
ymeine: You are testing `every` for a suitable use case which is to break from a loop, but one of its more generic and first place use cases is to return a meaningful (boolean) value. Please test it.

tjw02:
Thanks, I have made the further changes

test/aria/utils/ArrayTestCase.js
ymeine: I just posted my review when you did your last changes I think. This is a bit out of date, but still I think repeating the name of the tested function (`every`) all the time is not necessary.

tjw02:
Changes to the ArrayTestCase have been committed. I have removed the duplicate test for false, added the assert that the results are as expected, and updated the comments.

tjw02:
Fake commit to retrigger tests

tjw02:
Hi, the test failure above wasn't related to my change, but the above fake commits don't seem to retrigger the tests. Any ideas?

ymeine:
Sorry for the delay, your changes are all fine. Regarding the build, indeed it's not due to any of your actions, it doesn't matter as long as the other triggered builds are passing (sometimes it happens some tests fail exceptionally, in a random fashion). I'm gonna integrate your changes. Thanks a lot for your PR!! :smiley:

tjw02:
Thanks ymeine!

PR #1687, opened 2016-09-16
Removing test suites

divdavem:
This PR is supposed to replace #1681


divdavem:

PR #1688, opened 2016-09-16
IOXDRTestCase no longer uses pipes.yahooapis.com to test XDR

divdavem:
We now use `echo.ariatemplates.com`, as `pipes.yahooapis.com` is no longer available.


divdavem:

PR #1689, opened 2016-09-20
Adding waiHasPopup on widgets extending ActionWidget

divdavem:
This PR adds the `waiHasPopup` property on all widgets which extend `ActionWidget` (`@aria:Button`, `@aria:Link`, ...).


divdavem:

PR #1693, opened 2016-09-29
Removing `aria:parent` properties when they are no longer useful

divdavem:
This commit adds some checks to remove `aria:parent` properties on an object when that object no longer has any recursive listener and no longer has any parent in `aria:parent`. This fixes some memory leaks which can occur in case some parts of the data model are reused inside new containers and old containers are no longer used but still pointing to the reused parts of the data model (so the `aria:parent` properties on the reused parts of the data model still pointed to the old objects, even though there was no recursive listener any more in that old part of the data model).

### DIFF
--- a/src/aria/utils/Json.js
+++ b/src/aria/utils/Json.js
@@ -72,6 +72,9 @@ var ariaUtilsObject = require("./Object");
                     break;
                 }
             }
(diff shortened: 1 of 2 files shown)


divdavem:

PR #1697, opened 2016-10-03
fix/test test.aria.templates.customization.CustomTemplateTestCase

divdavem:
This commit fixes a random issue happening at the end of `CustomTemplateTestCase`:
restoring the environment at the end of the test triggers an XHR which can cause issues on the following tests in the test campaign.


divdavem:

PR #1698, opened 2016-10-04
feat #1698 Improved the DomNavigationManager & added a aria-hidden manager

ymeine:
Also:

- added a utility function to add click steps in JAWS tests
- added aria-hidden attributes on elements of the test runner 'test/test.htm', for safety and consistency

### DIFF
--- a/src/aria/popups/PopupNavigationManager.js
+++ b/src/aria/popups/PopupNavigationManager.js
@@ -1,17 +1,246 @@
-// -----------------------------------------------------------------------------
+/*
+ * Copyright 2016 Amadeus s.a.s.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 
-var Aria = require("../Aria");
-var ariaUtilsArray = require("../utils/Array");
-var ariaTemplatesNavigationManager = require("../templates/NavigationManager");
-var ariaUtilsEvent = require("../utils/Event");
+
+
+////////////////////////////////////////////////////////////////////////////////
+// Imports
+////////////////////////////////////////////////////////////////////////////////
+
+var Aria = require('../Aria');
+
+var ariaUtilsType = require('../utils/Type');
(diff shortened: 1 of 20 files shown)


src/aria/popups/PopupNavigationManager.js
```diff
@@ -433,5 +642,236 @@ function ViewportNavigationInterceptor() {
...
+    hideRequestsCount--;
+
+    if (hideRequestsCount === 0) {
+        element.removeAttribute('aria-hidden', 'true');
+        showed = true;
+    }
+
+    // --------------------------------------------------------- output & return
+
+    if (hideRequestsCount === 0) {
```
divdavem: The same condition is tested just above. Please factorize this code.

src/aria/popups/PopupNavigationManager.js
```diff
@@ -433,5 +642,236 @@ function ViewportNavigationInterceptor() {
...
+    var hideRequestsCount = element.getAttribute(attributeName);
+    if (hideRequestsCount == null) {
+        hideRequestsCount = 0;
+    }
+
+    // -------------------------------------------------------------- processing
+
+    var hidden = false;
+
+    if (hideRequestsCount === 0) {
```
divdavem: Under normal circumstances (if no one changes the attribute outside of your `hide` and `show` functions), `hideRequestCount` will never be 0 unless it was null and set to 0 just before... so I think you can simply merge the two `if` blocks together.

src/aria/popups/PopupNavigationManager.js
ymeine: For me they were a bit semantically distinct, hence the fact I did not merge them, but I can do it.

src/aria/popups/PopupNavigationManager.js
ymeine: The two blocks don't have the same meaning, I prefer keeping an integer value for `hideRequestsCount`, independently of how the integer value is ensured. If ever I want to keep the attribute in the element even when reaching `0`, that way I can only modify the code in the method `show`, here in `hide` it will still work because this is more robust and respects separation of concerns.

src/aria/popups/PopupNavigationManager.js
```diff
@@ -433,5 +642,236 @@ function ViewportNavigationInterceptor() {
...
+
+    return reverter;
+};
+
+/**
+ * Show back all elements around the given element, by applying the same traversal as for hideOthers.
+ *
+ * @param {HTMLElement} element The reference element
+ */
+prototype.showOthers = function(element) {
```
ymeine: Actually this can be used outside of this framework (and I know it is, even though here I preferred the method of the revert function). That brings two different ways to achieve the same thing theoretically.

But I just realized that indeed things start to differ when DOM elements are being added in the meantime, and there is a problem if some code requests hiding those elements. To avoid such things I would have to bind a concept of ids to the requests, thing I wanted to avoid by using a counter instead.

PR #1703, opened 2016-10-07
Removing alert when logging after aria.core.Log is disposed

divdavem:
When aria.core.Log is disposed, the previous code was replacing the error logging function by a function which displayed the error using alert. The alert could sometimes be triggered during the tests in IE 7 and
blocked the following tests. This commit replaces it with logs on the console.


divdavem:

PR #1704, opened 2016-10-07
Disabling ViewportHugeShrinkTestCase

divdavem:
`test.aria.widgets.container.dialog.viewportResize.ViewportHugeShrinkTestCase` was present in the repository without being part of a test suite, and has not been run during the release test campaign for a long time. It fails in all browsers except PhantomJS, but PhantomJS is known not to properly handle the size of the viewport (and this test is related to it).

More investigation should be done on this test to either fix it or remove it.


divdavem:

PR #1707, opened 2016-10-10
fix/test #1707 test.aria.core.CSSPrefixTestCase

divdavem:
This PR fixes `CSSPrefixTestCase` on Firefox 49 by changing the order in which prefixes are tested. `Moz` is now tested before `Webkit`, so that, on Firefox, `Moz` is chosen rather than `Webkit` (now that both seem to be supported on Firefox 49).


divdavem:

PR #1708, opened 2016-10-10
Using Firefox instead of PhantomJS to run TooltipRobotTestCase on travis-ci

divdavem:
Using Firefox instead of PhantomJS to run TooltipRobotTestCase on travis-ci as TooltipRobotTestCase often fails on PhantomJS (for an unknown reason).


divdavem:

PR #1709, opened 2016-10-11
fix #1709 Fixed aria.utils.DomNavigationManager for IE 7, 8, 9

ymeine:
Because of various API or syntax incompatibilities.


src/aria/utils/DomNavigationManager.js
```diff
@@ -688,6 +688,9 @@ function ElementNavigationInterceptor(element, loop) {
             // --------------------------------------------------- destructuring
 
             var target = event.target;
+            if (target == null) {
+                target = event.srcElement;
+            }
```
divdavem: I would have written this more simply on one line:

```js
var target = event.target || event.srcElement
```

PR #1711, opened 2016-10-11
Slowing down BindableRobotTestCase to make it more reliable

divdavem:
Depending on how fast the robot implementation is, `test.aria.html.textarea.bind.BindableRobotTestCase` and `test.aria.html.textinput.bind.BindableRobotTestCase` could fail in case the keys are typed quicker than the delay we use to simulate the type event.
This PR adds some pause between each character so that the corresponding type events do not combine several letters together.
Especially, this commit fixes both tests with the robot-server on Safari.


divdavem:

PR #1712, opened 2016-10-11
Improving reliability of DataModelOnChangeRobotTestCase

divdavem:
This PR improves the reliability of `DataModelOnChangeRobotTestCase` by waiting for the focus and adding a pause after typing. This fixes the test on Safari with robot-server.

### DIFF
--- a/test/aria/html/select/onchange/DataModelOnChangeRobotTestCase.js
+++ b/test/aria/html/select/onchange/DataModelOnChangeRobotTestCase.js
@@ -36,14 +36,25 @@ Aria.classDefinition({
 
             this.assertEquals(selectWidget.selectedIndex, 0, "The selected Index should be %2 but was %1");
 
-            this.synEvent.execute([["click", selectWidget], ["type", null, "[down][down][enter]\t"]], {
-                fn : this.afterChange,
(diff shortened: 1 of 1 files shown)


test/aria/html/select/onchange/DataModelOnChangeRobotTestCase.js
```diff
@@ -36,14 +36,19 @@ Aria.classDefinition({
 
             this.assertEquals(selectWidget.selectedIndex, 0, "The selected Index should be %2 but was %1");
 
-            this.synEvent.execute([["click", selectWidget], ["type", null, "[down][down][enter]\t"]], {
+            this.synEvent.execute([
+                ["click", selectWidget],
+                ["waitFocus", selectWidget],
+                ["type", null, "[down][down][enter]\t"],
+                ["pause", 1000]
```
benouat: 1 second seems to be huge.... Can't we use something a little less longer ?

divdavem: @benouat Thank you for your comment. I have changed the code to use `waitFor` instead.

PR #1714, opened 2016-10-13
Excluding failing tests

divdavem:
This commit creates a list of failing tests (for each browser), which will be excluded during the release campaign until they are individually investigated and fixed.


divdavem:

PR #1716, opened 2016-10-14
Fixing LinkDisabledTestCase on IE 8 (by making it a robot test)

divdavem:
Now that `@aria:Link` widgets really have the disabled attribute when they are disabled, it seems IE 8 raises an error when aria.utils.SynEvent tries to click on them. This commit fixes this issue by using the robot implementation of SynEvent (aria.jsunit.SynEvent) instead of aria.utils.SynEvent.


divdavem:

PR #1717, opened 2016-10-20
Using CORS instead of Flash by default for cross-domain requests

divdavem:
Now that CORS is largely implemented in browsers, this commit changes the default implementation of cross-domain requests to use CORS instead of the Flash-based implementation.

The Flash-based implementation can still be re-enabled by calling the `aria.core.IO.updateTransports` method in the following way:

```js
  aria.core.IO.updateTransports({
     'crossDomain' : 'aria.core.transport.XDR'
  });
```

Note that this change is NOT backward compatible. Applications relying on the Flash-based implementation of cross-domain requests should use the above lines of code or migrate to CORS.

### DIFF
--- a/src/aria/core/IO.js
+++ b/src/aria/core/IO.js
@@ -178,7 +178,7 @@ module.exports = Aria.classDefinition({
          * @type String class name and path.
          * @private
          */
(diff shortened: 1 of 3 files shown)


divdavem:

PR #1718, opened 2016-10-21
aria.utils.Bridge now uses aria.utils.FrameATLoader

divdavem:
aria.utils.Bridge now relies on aria.utils.FrameATLoader to load Aria Templates in the popup window instead of relying on its own implementation (which reduces duplication in the framework).

There is now a new `crossDomain` option in `aria.utils.FrameATLoader` to use document.write instead of loading `FrameATLoaderHTML.html`, which makes `FrameATLoader` work in case Aria Templates is loaded from a different origin (such as [cdn.ariatemplates.com](http://cdn.ariatemplates.com)).
However, when this option is set to `true`, changing the hash in the popup causes issues on Firefox 3 and 11, so it is `false` by default.

Calls to `document.write` to load a script, are now removed, as these calls can be blocked on Chrome in some cases, as documented in the following document:
https://developers.google.com/web/updates/2016/08/removing-document-write

This commit fixes `BridgeTestCase` on IE 9.
It also fixes the crash which occured when closing Aria Templates debug tools (due to an endless recursive call).


divdavem:

PR #1719, opened 2016-10-25
Skip updating manually hidden dialogs on viewport resize

divdavem:
If a dialog is manually hidden (by setting its `display` CSS property to `none` and by updating the corresponding `isOpen` boolean on the popup object), its size will no longer be updated on viewport resize.
This avoids setting the size of the popup to 0 (as offsetWidth and offsetHeight can return 0 for elements which are not visible on the screen).


divdavem:

PR #1720, opened 2016-10-25
Updating dependencies

divdavem:
This commit updates the dependencies of Aria Templates.

Before integrating it, we should:
- [x] wait for a new release of UglifyJS which includes the fix for https://github.com/mishoo/UglifyJS2/issues/1345
- [x] update the version of UglifyJS used in at-packager
- [x] make sure the screw_ie8 parameter defaults to false in at-packager (even if the default value changed in version 2.7.0 of UglifyJS, cf https://github.com/mishoo/UglifyJS2/commit/02c638209ee22816b1324ff0c0f47b27db1336af)
- [x] publish a new version of at-packager
- [x] update the version of at-packager in package.json


divdavem:

divdavem:
The travis build failure is caused by the following bug in version 1.7.4 of UglifyJS: https://github.com/mishoo/UglifyJS2/issues/1345

PR #1722, opened 2016-11-07
Removing the "tab" role

divdavem:
Blind users do not like the "tab" role because, just as the "application" role, it prevents them from using their navigation shortcuts. This commit simply removes the role attribute for tabs, and also adds the href attribute on the anchors so that tabs are recognized as links.


divdavem:

PR #1725, opened 2016-11-14
Removing role="button" and adding a tabindex on dropdown icons

divdavem:
This commit removes the `role="button"` attribute and adds a `tabindex` on the dropdown icon of the widgets which extend `DropDownTextInput` (`DatePicker`, `MultiSelect`...), so that those icons are not included in the list of buttons displayed by screen readers and become focusable by tab.


divdavem:

PR #1730, opened 2016-11-21
Adding a focusable element at the top of the TabPanel widget

divdavem:
This focusable element is only added when `waiAria` is true, and has the same `aria-labelledby` attribute as the root element of the tab panel, so that, when focusing the tab panel, the title of the currently selected tab is read by screen readers.


divdavem:

PR #1731, opened 2016-11-22
Adding the focusableWhenDisabled parameter on aria:Button widgets

divdavem:
Blind users often navigate in a page with tab, which means they cannot notice disabled widgets, which are not focusable.
When those widgets are essential to understand the page, this is a problem, especially for buttons. This commit allows important buttons to be focusable when they are disabled, by setting the `focusableWhenDisabled` parameter to `true`.


divdavem:

PR #1732, opened 2016-11-24
Removing specific versions of Chrome and Firefox in testConfigBuilder.js

divdavem:
This avoids having to update the repository for each new version of those browsers.


divdavem:

PR #1733, opened 2016-11-24
Updating Firefox to version 50 (from 48) for travis tests

divdavem:


divdavem:
@ymeine Thank you for your review! You know that, sometimes, just a missing comma (1 character change) can make a huge difference!

PR #1734, opened 2016-11-29
Injecting DomEventWrapper in TextInput:onclick

divdavem:
Before this commit, the `onclick` event handler of the TextInput-based widgets did not receive any event object. This commit makes sure a `DomEventWrapper` object is passed to the event handler, as it is done in other widgets.

Note that this PR implements a subset of #1638


divdavem:

PR #1736, opened 2016-12-02
Changing implementation of interface wrappers to avoid leaks

divdavem:
If an object for which an interface wrapper has been created is never disposed through its $dispose method, the garbage collector can never collect it because the corresponding interface keeps a reference to it.

For example, assuming the `x.Y` class impements the `x.IY` interface, the following code creates a leak:

```js
 for (var i = 0; i < 100000; i++) {
    var myObj = new x.Y();
    myObj.$interface("x.IY");
    // The leak would not appear if myObj was disposed like this:
    // myObj.$dipsose();
 }
```

This commit avoids creating a leak in such a case by using a `WeakMap` in the interface in order to keep a reference to the underlying object without preventing the garbage collector from collecting the underlying
object and interface wrapper if there is no other reference to them anywhere.
(in case the browser does not support `WeakMap`, an equivalent alternative implementation is used)

### DIFF
--- a/src/aria/core/Interfaces.js
+++ b/src/aria/core/Interfaces.js
@@ -41,6 +41,28 @@ var Aria = require("../Aria");
         return key;
     };
 
+    var __weakMap = Aria.$global.WeakMap || (function () {
+        // mini weak map implementation for our needs
+        var WeakMap = function () {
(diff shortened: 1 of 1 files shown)


divdavem:

PR #1738, opened 2016-12-05
easypackage and atbuild issues with npm3

divdavem:
This PR fixes issues with the `easypackage` and `atbuild` grunt tasks when using npm3 (with its flat structure).


divdavem:

PR #1739, opened 2016-12-08
Memory leak when calling $getElementById or $getChild on a template

divdavem:
Each time a call to `$getElementById` or `$getChild` is done on a template, a new `DomElementWrapper` object is created and, until this commit, the wrapper was also added to an array in the template context, with the only purpose of calling `$dispose` on all wrapper elements when the template is fully refreshed or disposed. If this does not happen (e.g. if there are only partial refreshes), and many calls to the `$getElementById` and `$getChild` functions are done, this creates a leak. With this commit, wrappers are no longer stored on the template context, it is up to the application to dispose them (if needed).

### DIFF
--- a/src/aria/templates/DomElementWrapper.js
+++ b/src/aria/templates/DomElementWrapper.js
@@ -291,7 +291,6 @@ module.exports = Aria.classDefinition({
          * @private
          */
         this._dispose = function () {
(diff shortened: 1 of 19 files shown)


divdavem:
@benouat and @fbasso Thank you for your review!

PR #1740, opened 2016-12-13
Fix to error logging in the JsonValidator validateCfg method

tjw02:
After speaking with a member of the aria templates team, this change to the JsonValidator is to allow the logging to also include the properties that are in error. Up til now these were being ignored as they are in an errors list, which is being ignored in the DefaultAppender. The member of the AT team suggested this as an implementation.


divdavem:
@tjw02 Thank you for your pull request! That's a nice and useful contribution!

PR #1741, opened 2016-12-14
Adding the waiOpenMsg property on the aria:Dialog widget

divdavem:
The new `waiOpenMsg` property allows to specify a message to be read by screen readers when the dialog is opened.


divdavem:

PR #1742, opened 2016-12-15
Regression in linkItfWrappers after refactoring interface wrappers

divdavem:
Commit ff7921e66b5a12e6de8e31ca55be79c0c93a4fb0 from #1736 introduced a regression in the `linkItfWrappers` function on modern browsers (which have an implementation of `WeakMap`).
As a consequence, test.aria.modules.moduleReload.ModuleReload1TestCase and test.aria.modules.moduleReload.ModuleReload2TestCase fail on those browsers.
This pull request fixes this regression.


divdavem:

PR #1743, opened 2016-12-15
Missing call to $dispose on DomElementWrapper

divdavem:
Following commit 925c9078cfea9cdaccab764330e217f18696fad6, the call to `$dispose` on `DomElementWrapper` objects returned by `$getElementById` is no longer done automatically. This commit adds a missing call to `$dispose` on a `DomElementWrapper` instance in a Jaws test case.


divdavem:

PR #1747, opened 2017-01-11
Dropdown icons no longer focusable when the widget is disabled

divdavem:
This PR removes the "tabindex" attribute and adds the "aria-hidden" attribute on the dropdown icon of widgets when they are disabled.


divdavem:

PR #1748, opened 2017-01-16
Avoid refreshing dialog content after a resize

divdavem:
This PR removes the refresh which happens after resizing a dialog.


divdavem:

PR #1749, opened 2017-01-20
Adding the onblur event on the Select widget

divdavem:
This commit also adds some fixes to make sure there is no blur event when clicking on the dropdown icon or when clicking on an option in the list.


divdavem:

PR #1750, opened 2017-01-20
Assertion methods for error paths

pschichtel:
For complete test coverage, it is necessary to test error paths as well, but AT does not yet provide simple assert* methods to invoke a code section and check if errors are thrown and if the correct errors are thrown.

### DIFF
--- a/src/aria/jsunit/Assert.js
+++ b/src/aria/jsunit/Assert.js
@@ -413,6 +413,36 @@ var ariaUtilsJson = require("../utils/Json");
             },
 
             /**
+             * Asserts that an error is thrown in the given block (function)
+             * @param {Function} block the function to invoke and detect errors in
+             * @param {String} optMsg optional message to add to the failure description
+             * @return {*}
+             */
+            assertError: function(block, optMsg) {
+                var ex = null;
+                try {
+                    block.apply(this);
+                } catch (e) {
+                    ex = e;
+                }
+                this.assertNotNull(ex, optMsg || "No error has been thrown!");
+                return ex;
+            },
+
+            /**
+             * Asserts that a certain type of error / exception is thrown in the given block (function)
+             * @param {Object} proto the base type of the expected error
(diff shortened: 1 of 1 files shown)


pschichtel:
@divdavem as you seem to be the only active commiter recently I'm asking you directly: Are PRs still being processed? This PR hasn't been touched since opening it over 2 years ago. If there is no interest in the proposed changes I'd be fine with closing it.

divdavem:
@pschichtel Thank you for your PR. We are sorry not to have answered before.

It is true that it is important to test error paths as well. In the Aria Templates framework, we mostly reported errors (at least at the beginning) with `this.$logError` instead of throwing an exception and those errors can be checked with `assertErrorInLogs`. There are probably also cases where we throw exceptions, and those cases may not be fully tested.

However, the Aria Templates framework is now mostly in maintenance mode and not advised for new applications as there are more modern and better javascript templating frameworks available nowadays.

Do you still use Aria Templates yourself? Would you use this feature if we merged this PR?
If it is not the case, I think we can close the PR because, as far as I know, there is not much new code written with Aria Templates that would require writing new tests to check those error cases.

In case you think it is still useful to merge the PR, I have a small remark about the code: it is possible in javascript to throw a `null` error, and in that case `assertError` would not behave properly with your implementation as far as I can see, because throwing null would be considered as not throwing anything. I admit this is a very rare edge case, but I think it is still interesting to mention it.

Again thank you for your contribution and sorry for the delay to give you an answer!

pschichtel:
I really don't have any use for aria templates. This PR is from the time when I was working for Amadeus. I implemented these two functions in the project and we thought why not upstream them?

So feel free to close this, I was just looking for closure :). Good catch about the `throw null`, I would fix it if there was interest in merging this.

divdavem:
Thank you for your answer! It is a good thing, indeed, to think about others and share reusable code.
It could have been useful if we had merged it on time. Sorry!
I am closing the PR now! Thank you!

PR #1753, opened 2017-02-03
Fixing issues in tests following #1749

divdavem:
This PR fixes test issues which are happening following commit 8432f5b24f516bfc1361181d98535736c5b9fd47 from #1749:
- `SelectOnBlurSimpleRobotTestCase` fails on Safari because a `<select>` in Safari (on Mac) behaves differently than in other browsers (on Windows): pressing the down arrow key opens the dropdown instead of only changing the selection, as it is expected in the test. This test is now excluded from the test campaign for Safari.

- `OpenDropDownFromButtonTestCase` fails on IE 8 because SynEvents does not faithfully reproduces the behavior of IE 8. This test was replaced by a robot test which correctly tests what is really happening on IE 8.


divdavem:

PR #1756, opened 2017-02-22
Adding Romanian date resources

divdavem:
This PR adds partial support for the Romanian language, which can be used to make `aria.utils.Date.format` return formatted dates in Romanian.


divdavem:

PR #1757, opened 2017-02-22
feat/test Adding test for #1755

divdavem:
This PR adds a test for #1755.


divdavem:

PR #1758, opened 2017-02-27
Keep existing log appenders in aria.jsunit.Assert

divdavem:
Before this PR, creating an instance of `aria.jsunit.Assert` (the base class for most tests) cleared all appenders in `aria.core.Log`. Now, this class only manages its own appender to collect errors during tests and no longer impacts any other appender.
See also #590.

Also, a second commit in the PR removes the deprecation messages for `unloadClass` and `getFilename` as those methods are used in the framework itself (especially in base test classes but also in some other places) and the messages can then be confusing for an application developer which does not call those methods directly, and those messages are more visible in test logs if appenders are not cleared (and they are no longer cleared in the `Assert` class).

### DIFF
--- a/src/aria/core/Cache.js
+++ b/src/aria/core/Cache.js
@@ -36,9 +36,7 @@ module.exports = Aria.classDefinition({
             "CML" : ".cml",
             "TXT" : ".tpl.txt",
             "RES" : ".js"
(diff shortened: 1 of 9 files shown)


divdavem:

PR #1759, opened 2017-03-06
Beans pre-compilation

divdavem:
This PR adds support for beans pre-compilation and optimizes a bit `aria.core.JsonValidator` to avoid generating twice the same fast-normalization function when a bean can simply reuse the fast-normalization function of its parent.


divdavem:
Note that this PR depends on ariatemplates/atpackager#26

divdavem:

PR #1760, opened 2017-03-07
Fixes and improvements for beans pre-compilation

divdavem:
This PR contains a set of fixes and improvements linked to beans pre-compilation, introduced in #1759:
- the `ATDebug` flag when compiling beans is now removed because it also caused all templates to be compiled in debug mode (with line numbers)
- there are new checks in `BeanExtractor` to be sure that the configuration in the `JsonValidator` is correct for the extraction with the given options to be successful
- an issue with default values which were not correctly normalized in non debug mode is now fixed
- the value returned by the generated `$getDefault` functions is no longer fully normalized, but it is expanded in `$fastNorm` in order to reduce the size of compiled beans
- beans which are not used in the fast normalization process are now discarded by the bean extractor when `onlyFastNorm` is true, in order to reduce the size of the resulting file
- functions returning typical default values (`{}`, `[]`, and `null`) are now cached to avoid generating them multiple times and to reduce the size of compiled beans
- the generated fast normalization functions now call the fast normalization function generated for the parent bean when it is useful and possible, in order to reduce the size of compiled beans
- a consistency issue between fast and slow normalization is now fixed by removing early normalization of default values. Default values are no longer normalized at the time a bean is pre-processed. At the time a default value is used, its copy is normalized in the context where it is used. This can change the result of normalization in case a bean overrides the default value of some of its parent properties and inherits from the default value of its parent (which is no longer normalized).

### DIFF
--- a/build/grunt-config/config-atpackager-prod.js
+++ b/build/grunt-config/config-atpackager-prod.js
@@ -40,7 +40,6 @@ module.exports = function (grunt) {
 
     grunt.config.set('atpackager.prod', {
         options : {
-            ATDebug : true,
             ATBootstrapFile : packagingSettings.bootstrap.bootstrapFileName,
             sourceDirectories : packagingSettings.prod.sourceDirectories,
             sourceFiles : ['**/*', '!aria/node.js', '!' + packagingSettings.bootstrap.bootstrapFileName],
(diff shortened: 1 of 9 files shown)


divdavem:

PR #1761, opened 2017-03-10
fix aria:Button had its title attribute on the span instead of button

divdavem:
This PR fixes an issue with the aria:Button widget with screen readers: the fact that the title attribute was not set on the button tag but on the containing span was preventing Jaws from reading the tooltip.


divdavem:

PR #1762, opened 2017-03-16
Excluding failing tests from campaign

divdavem:
The following tests are now excluded from the test campaign because they often fail (those tests should be rewritten in a more reliable way):

on Firefox 3:
 test.aria.utils.HistoryTestCase on Firefox 3
on IE 11 with JAWS:
 test.aria.widgets.wai.errorlist.binding.ErrorListBindingJawsTestCase


divdavem:

PR #1764, opened 2017-03-29
Allow to focus the link inside the aria:Tab widget

divdavem:


divdavem:

PR #1766, opened 2017-05-31
fix Text input error tooltip wrong position when block is set to true

divdavem:


divdavem:

PR #1767, opened 2017-06-02
Adding Slovak date resources

divdavem:
This PR adds partial support for the Slovak language, which can be used to make aria.utils.Date.format return formatted dates in Slovak.

cf #1756 for the last added language (date resources only)


divdavem:

PR #1770, opened 2017-07-20
fix/test rewriting TransitioncaseRobotTestCase

divdavem:
test.aria.templates.issue348.transition.TransitioncaseRobotTestCase was failing because of a timeout. This test was rewritten to be cleaner.


divdavem:

PR #1771, opened 2017-08-18
Adding Bulgarian date resources

divdavem:
This commit adds partial support for the Bulgarian language, which can be used to make aria.utils.Date.format return formatted dates in Bulgarian.

cf #1767 for the last added language (date resources only)


divdavem:

PR #1772, opened 2017-09-04
dev #1772 Adding package-json.lock to ensure installed dependencies are the same for everyone

ymeine:


divdavem:
I am closing the PR as this change is included in #1773

PR #1774, opened 2017-09-13
Improving the reliability of Jaws tests

divdavem:
Sometimes, Jaws tests are executed but the history retrieval fails (probably because of a timing issue).
This PR adds some more pauses and repeats some actions multiple times when copy/pasting the history to improve the reliability of Jaws tests.


divdavem:
This PR is not really needed anymore, especially because of the improvements brought by #1811.

PR #1775, opened 2017-09-13
New `ariaLive` parameter on the `@aria:ErrorList` widget

divdavem:
This new `ariaLive` parameter allows to specify whether `role="alert"` and `aria-live="assertive"` are included (`true`) or omitted (`false`) on the container element of the `@aria:ErrorList` widget (defaults to `true`, only used if `waiAria` is `true`)


divdavem:
@fbasso: Thank you for your review!

PR #1776, opened 2017-09-18
Adding Mongolian date resources

divdavem:
This commit adds partial support for the Mongolian language, which can be used to make `aria.utils.Date.format` return formatted dates in Mongolian.

cf #1771 for the last added language (date resources only)


divdavem:

PR #1780, opened 2017-10-02
fix wrong computation of modal mask size

divdavem:


src/aria/popups/Popup.js
```diff
@@ -731,11 +750,16 @@ module.exports = Aria.classDefinition({
                     this._containerOverflow = popupContainer.changeContainerOverflow("hidden");
                 }
 
-                var containerSize = popupContainer.getScrollSize();
+                // temporarily hides the modal mask so that it does not have an influence the size of the container
```
ymeine: typo: "on" the size

test/aria/widgets/container/dialog/modalMaskSize/DialogModalMaskSize.tpl
```diff
@@ -0,0 +1,176 @@
...
+            bind: {
+                value: {
+                    to: "visible",
+                    inside: data
+                }
+            }
+        } /}
+        <div style="width: 100px; height: 2000px; background-color:blue;"></div>
+        <input {id "myInput"/}>
+        {@aria:CheckBox {
```
ymeine: You have an identical Checkbox just above. Is it a mistake or done on purpose?

test/aria/widgets/container/dialog/modalMaskSize/DialogModalMaskSizeTestCase.js
```diff
@@ -0,0 +1,101 @@
...
+    },
+    $prototype : {
+        runTemplateTest : function () {
+            var self = this;
+            var testWindow = self.testWindow;
+            var document = testWindow.document;
+            var body = document.body;
+            var documentElement = document.documentElement;
+            var jsonUtils = testWindow.aria.utils.Json;
+            var domUtils = testWindow.aria.utils.Dom;
```
ymeine: Why not using the dependencies from the current context instead of using the ones from the test? (which are theoretically not ensured to be available, even if in this case we know they'll be)

I you keep your code like this, remove the `aria.utils.Json` dependency from the `$dependencies` field. If you change it instead, don't forget to add `aria.utils.Dom` to it.

test/aria/widgets/container/dialog/modalMaskSize/DialogModalMaskSize.tpl
```diff
@@ -0,0 +1,176 @@
...
+        {@aria:NumberField {
+            label: "boxWidth",
+            bind: {
+                value: {
+                    to: "boxWidth",
+                    inside: data
+                }
+            }
+        } /}<br>
+        {@aria:NumberField {
```
ymeine: Are all those fields above useful for manual testing? Otherwise (and only otherwise) I would say to remove them since they are not used in the automated test.

test/aria/widgets/container/dialog/modalMaskSize/DialogModalMaskSize.tpl
divdavem: It is done on purpose to be able to trigger the display of the dialog whether the page is scrolled or not (there is a big div separating both checkboxes).

test/aria/widgets/container/dialog/modalMaskSize/DialogModalMaskSize.tpl
divdavem: Yes, those fields are useful for manual testing.

src/aria/popups/Popup.js
divdavem: Thank you for catching this mistake. I will fix it.

test/aria/widgets/container/dialog/modalMaskSize/DialogModalMaskSizeTestCase.js
divdavem: I have now added `aria.utils.Json` and `aria.utils.Dom` to the `$dependencies` array of the template to make sure that those dependencies are included in the test window, even though this is not really mandatory, because those classes are used in the template loading process (so we know that they will already be loaded).

divdavem: I have also removed `aria.utils.Json` from the `$dependencies` field of the test.

divdavem:

PR #1784, opened 2017-11-13
activateSort in aria:List not taken into account with default template

divdavem:
The `activateSort` property of the `@aria:List` widget was not taken into account with the default list template. There was also an issue with the way the arrow up and arrow down keys were working when the list is sorted. This PR fixes those problems.


divdavem:
@ymeine Thank you for your review and your nice comment!

PR #1785, opened 2017-11-22
fix overlay z-index computation (now takes child elements into account)

divdavem:
This commit fixes a regression caused by commit 45f686eaa6e018b6eaa4b91db74a774b20b65028, by uncommenting the code already written by @ymeine to take into account child elements when computing the z-index to apply on an element.


divdavem:
@ymeine Thank you for your review, and thank you even more for having written the code initially (so that the fix is easy for me now!).

divdavem:

PR #1786, opened 2017-11-30
fixing issue while installing "awesome" on travis-ci

divdavem:
It seems that the `awesome` package raises random issues while being installed (cf https://travis-ci.org/ariatemplates/ariatemplates/builds/309427769 and https://travis-ci.org/divdavem/ariatemplates/builds/309427661 one installation is failing, the other is successful, both from the same commit).
Let's remove it and disable the test that depends on it to make our build more reliable (also, that test was not always working reliably).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,10 @@ cache:
     - node_modules
 addons:
   firefox: "55.0"
-  apt:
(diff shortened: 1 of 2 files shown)


divdavem:

PR #1788, opened 2018-01-29
fix Selectbox issue with helptext on IE 11

divdavem:
This PR fixes test.aria.widgets.form.selectbox.SelectboxTestCase on IE 11.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ node_modules
 
 /build/target/
 /test/target/
+/test/logs/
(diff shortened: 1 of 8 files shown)


divdavem:
@ymeine Thank you for your review! As discussed, this change breaks `test.aria.widgets.form.multiautocomplete.issue1076.focus.OnFocusRobotTestCase` on IE 11. I will fix this issue before integrating this PR.

PR #1789, opened 2018-02-01
fix radio button error when it is disposed while setting its value

divdavem:


divdavem:
Nothing revolutionary (even if it is #1789)!

divdavem:

divdavem:
PTR 14581672

PR #1790, opened 2018-02-16
Adding restoreFocusOnClose parameter in Dialog widget

divdavem:
This commit adds the "restoreFocusOnClose" parameter on the Dialog widget, which is true by default (to keep the same behavior as before which restores the focus to the element that was focused before the dialog was opened when the dialog is closed) but which can now also be set to false (either directly in the configuration, or through a binding) to disable the automatic restoration of the focus. This is especially useful if the application manages the focus itself when the dialog is closed and does not want the framework to interfere.
This parameter is ignored for non-modal dialogs (the focus is never restored in that case).


divdavem:

PR #1791, opened 2018-03-05
Fix: memory leak introduced by the call eval instead of JSON.parse th…

MS-elug:
…at create global variable attached to the GC root node and never disposed

PTR 14797882-Memory Leaks Fixes

Gist to reproduce the issue:
https://gist.github.com/MS-elug/7f275e8d0dd6f2f6d5f2482632632ef7

### DIFF
--- a/src/aria/utils/Json.js
+++ b/src/aria/utils/Json.js
@@ -377,7 +377,8 @@ var ariaUtilsObject = require("./Object");
              */
             SPLICE : 3,
 
-            // ERROR MESSAGES:
+            // MESSAGES:
+            INVALID_JSON_SYNTAX : "Cannot parse string with native JSON.parse. Defaulting to eval.\ncontext: %1\nJSON string:\n%2",
             INVALID_JSON_CONTENT : "An error occured while loading an invalid JSON content:\ncontext: %1\nJSON content:\n%2",
             NOT_OF_SPECIFIED_DH_TYPE : "Invalid data holder type: expected to be an object or an array. Data holder provided: ",
             INVALID_SPLICE_PARAMETERS : "Invalid splice parameters.",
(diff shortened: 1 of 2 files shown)


src/aria/utils/Json.js
```diff
@@ -835,21 +835,47 @@ var ariaUtilsObject = require("./Object");
...
              */
-            load : function (str, ctxt, errMsg) {
+            load: function (str, ctxt, errMsg) {
+                var JSON = Aria.$global.JSON;
+                if (typeof JSON !== "undefined" && typeof JSON.parse === "function") {
+                    try {
+                        return JSON.parse(str);
+                    } catch (ex) {
+                        // Fallback to eval
+                        this.$logWarn("Cannot parse string with native JSON.parse. Defaulting to eval.\ncontext: %1\nJSON string:\n%2", [ctxt, str]);
```
divdavem: Please move this warning string to `$statics` (just like other error/warning messages).

src/aria/utils/Json.js
```diff
@@ -835,21 +835,47 @@ var ariaUtilsObject = require("./Object");
...
+             * aria.utils.Json.INVALID_JSON_CONTENT
+             * @return {Object} 
+             * @private
+             */
+            _loadWithEval: function (str, ctxt, errMsg) {
                 var res = null;
+
+                if (ctxt && ctxt.$classpath) {
+                    ctxt = ctxt.$classpath;
+                }
```
divdavem: I think this code should be in the `load` function, so that the warning also uses the `$classpath` property.

src/aria/utils/Json.js
```diff
@@ -835,21 +835,47 @@ var ariaUtilsObject = require("./Object");
...
+                if (!errMsg) {
+                    errMsg = this.INVALID_JSON_CONTENT;
+                }
```
divdavem: This could stay in the catch block.

src/aria/utils/json/JsonSerializer.js
```diff
@@ -580,17 +580,41 @@ module.exports = Aria.classDefinition({
...
-            parse : function (string) {
+            parse: function (string) {
                 var text = String(string);
+                var JSON = Aria.$global.JSON;
+                if (typeof JSON !== "undefined" && typeof JSON.parse === "function") {
+                    try {
+                        return JSON.parse(text);
+                    } catch (ex) {
+                        // Fallback to eval
+                        this.$logWarn("Cannot parse string with native JSON.parse. Defaulting to eval.\nJSON string:\n%1", [text]);
```
divdavem: Please move this warning string to $statics (just like other error/warning messages).

PR #1795, opened 2018-04-30
fix MultiSelect error when it is disposed while setting its value

divdavem:


divdavem:

PR #1796, opened 2018-04-30
fix/test missing calls to parent destructor in some test classes

divdavem:
These issues resulted in some error messages in the test logs:

``
  [Aria] Error: the $destructor of x was not called in y
``


divdavem:

PR #1797, opened 2018-05-09
fix JAWS reading dialog title when closing dropdown widget

divdavem:
This PR contains a workaround for an issue with the JAWS screen reader: JAWS reads the dialog title when closing a dropdown widget that is inside the dialog, whereas it should not.

As investigated by @ymeine:

> - the issue comes from the fact the dropdown's element is physically outside the dialog, the focus is moved to it, and when closing the dropdown the focus is put back to the dialog, making JAWS announce it again
> - we can't put the dropdown's element inside the dialog because of other limitations (possible overflow styling on parent elements)
> - the right way to do is to use the WAI-ARIA attribute `aria-owns` to restore the logical structure of elements but JAWS doesn't support it
> - the workaround is to remove the `role=dialog` attribute on the Dialog's root element as long as a dropdown logically located inside it is open, and restore it short after closing this dropdown

The workaround implemented in this PR unfortunately only delays the issue because when the user moves the focus to another item inside the dialog, JAWS will detect the `role=dialog` attribute that was restored on the dialog and read the dialog title again!

### DIFF
--- a/src/aria/widgets/form/DropDownTrait.js
+++ b/src/aria/widgets/form/DropDownTrait.js
@@ -168,10 +168,52 @@ module.exports = Aria.classDefinition({
         },
 
         /**
+         * Removes role="dialog" from all parent elements and stores those elements in this._waiDialogRoleRemoved.
+         */
+        _removeDialogRole : function () {
+            if (!this._cfg.waiAria) {
+                return;
+            }
+            var waiDialogRoleRemoved = this._waiDialogRoleRemoved;
+            if (!waiDialogRoleRemoved) {
+                this._waiDialogRoleRemoved = waiDialogRoleRemoved = [];
+            }
+            var domElt = this.getDom().parentElement;
+            while (domElt) {
+                if (domElt.getAttribute("role") === "dialog") {
+                    domElt.removeAttribute("role");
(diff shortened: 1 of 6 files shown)


src/aria/widgets/form/DropDownTrait.js
```diff
@@ -167,11 +167,53 @@ module.exports = Aria.classDefinition({
...
+        /**
+         * Asynchronously restores role="dialog" on all elements in this._waiDialogRoleRemoved
+         * (and removes those elements from this array).
+         */
+        _restoreDialogRole : function () {
+            var waiDialogRoleRemoved = this._waiDialogRoleRemoved;
+            if (!waiDialogRoleRemoved || waiDialogRoleRemoved.length === 0) {
+                return;
+            }
+            setTimeout(function () {
```
benouat: Why do we need the 1s delay here ?

divdavem: I reused the same delay as in the work done by @ymeine.
I tried to use setTimeout with no delay, but in that case, it randomly fails, probably because the focus of the field, which is asynchronous in IE, is sometimes done after restoring `role="dialog"` on the parent element.

divdavem:

PR #1800, opened 2018-06-22
Excluding failing tests on IE 8, IE 9 and Edge

divdavem:


divdavem:

PR #1802, opened 2018-09-10
Moving aria/tester/index.html to aria-tester.html

divdavem:
This allows to publish to production the full `aria` folder without having to filter files inside it.


divdavem:

PR #1805, opened 2018-12-13
fix/test Fixing tests not working properly with large viewport

divdavem:
This PR fixes some tests that were not working properly when run with a large viewport because the viewport was able to fully display the content and the tests were expecting the viewport to have a scrollbar.
This commit fixes this by increasing the size of the content to make sure that a scrollbar is present.


divdavem:

PR #1806, opened 2018-12-14
fix/test Increasing pause after resizing the dialog in AbstractResizableDialogRobotBase

divdavem:


divdavem:

PR #1807, opened 2019-01-16
Adding autocomplete and name properties to text input widgets

divdavem:


divdavem:

PR #1808, opened 2019-01-22
fix JAWS reading dialog title when closing dropdown widget

divdavem:
This PR contains a better fix for the problem described in #1797.

When a dropdown widget with `waiAria:true` is inside a dialog with `role=dialog` the dropdown DOM element is no longer appended to the `<body>` but to the element that has `role=dialog`. Therefore, screen readers correctly understand that the dropdown is not outside the dialog and no longer read the dialog title when changing the focus after closing the dropdown.


divdavem:

PR #1809, opened 2019-01-25
Wrong week numbers displayed in aria:Calendar (and aria:DatePicker)

divdavem:
The calendar widget now uses `aria.utils.Date.getWeekNumber` (which is compliant with standard rules) to compute the week number, instead of considering that the first week of the year is the first full week.
For example, in 2019, if the first day of the week is Monday, week 1 should start on 31/12/2018 and not on 07/01/2019.

This commit extends `aria.utils.Date.getWeekNumber` to allow any day of the week to be specified as the first day of the week.
If Monday is selected, rules from ISO 8601 are applied (the first week of the year is the week containing January 4th).
Otherwise, the first week of the year is the week containing January 1st.

### DIFF
--- a/src/aria/utils/Date.js
+++ b/src/aria/utils/Date.js
@@ -188,8 +188,6 @@ var parseNextSteps = function(state, options) {
 
 };
 
-
-
 /**
  * This class contains utilities to manipulate Dates.
  */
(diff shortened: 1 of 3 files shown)


divdavem:

PR #1810, opened 2019-03-07
Screen reader reading previously focused autocomplete field on tab

divdavem:
This fixes an issue with JAWS, which reads the previously focused autocomplete field when pressing tab while the suggestions list is opened.
The problem seems to come from the `aria-activedescendant` attribute being removed too quickly (before the next field is focused).


divdavem:

PR #1819, opened 2019-06-18
feat/test Using puppeteer instead of phantomjs to run tests in travis-ci

divdavem:


scripts/ciCampaigns.js
```diff
@@ -49,6 +55,8 @@ process.on("exit", function () {
 });
 
 var options = {
+    "max-task-restarts": 3,
+    "task-restart-on-failure": true,
     "colors": true,
     "env": attester.config.readFile("package.json"),
     "phantomjs-instances": 0,
```
benouat: I think you can also remove this option.

divdavem:
@benouat Thank you for your review! I have made the change you requested!

PR #1821, opened 2019-06-20
fix/test HelpTextTestCase on IE 11

divdavem:


divdavem:
@ymeine Thank you for your review! Indeed, IE is sometimes (often?) mysterious.

PR #1826, opened 2020-04-27
Add a default invalid key for Bing maps

divdavem:
Bing maps is not supposed to be used without a key, and it even fails displaying a proper error message in that case.
This commit adds a default invalid key for quick tests allowing usage of the map in case a key is not provided as it should.


divdavem:

PR #1827, opened 2020-05-11
Allows to set escapeHtmlByDefault per template and per package

divdavem:


divdavem:

PR #1832, opened 2020-12-10
fix incorrect focus state in IE for aria:Tab widget

divdavem:
Clicking on `<td>` or `<span style="display:inline-block">` elements on IE focuses those elements whereas they are not focusable on other browsers (when `tabindex` is not specified).
This fix prevents this discrepancy between browsers from happening for the aria:Tab widget.


fbasso (approved):

divdavem:

PR #1838, opened 2021-03-30
ua-parser-js trailing comma

divdavem:
The new version of `ua-parser-js` (0.7.26) contains a trailing comma in an array. This is an issue with some (old) browsers.
This PR adds a reformatting step (included in the `JSConcat` builder of `atpackager`) so that the trailing comma is removed before `ua-parser-js` is included in Aria Templates (to avoid any problem).


fbasso (approved):
LGTM !

divdavem:

PR #1864, opened 2024-09-05
fix scrolling inside a popup on iOS 12 and 13 does not work

divdavem:
PTR 27563605


fbasso (approved):
Thanks for your fix @divdavem ! It seems clear and fine for me.

divdavem:

